PR #24, opened 2015-01-27
First working implementation

niXman:
Hi,

Please review the first working implementation.

For this macro-function:
```cpp
SQLPP_DECLARE_TABLE(
	(table)
	,
	(id, int, NOT_NULL)
	(c0, blob, NULL)
	(c1, bool, NOT_NULL)
	(c2, float, NULL)
	(c3, tinyint, NULL)
	(c4, smallint, NOT_NULL)
	(c5, int, NULL)
	(c6, bigint, NOT_NULL)
	(c7, text, NULL)
	(c8, varchar(16), NOT_NULL)
	(c9, varchar(32), NULL)
)
```
this code will be generated:
https://gist.github.com/niXman/553de6d36f8176a1ce58

Ok for merge?
Any proposals/objections?

### DIFF
--- a/examples/ppgen/ppgen.cpp
+++ b/examples/ppgen/ppgen.cpp
@@ -26,10 +26,28 @@
 
 #include <sqlpp11/ppgen.h>
 
+#if 0
 SQLPP_DECLARE_TABLE(
 	(table, ENGINE("InnoDB"), CHARACTER_SET("utf-8"))
 	,
 	(id, int, NOT_NULL, PRIMARY_KEY, AUTO_INCREMENT)
 	(name, varchar(64), NOT_NULL, INDEX("name_index"), DEFAULT("any name"))
 	(age, int, NOT_NULL, INDEX("age_index"), UNIQUE, COMMENT("some comments"))
+)
+#endif
+
+SQLPP_DECLARE_TABLE(
+	(table)
+	,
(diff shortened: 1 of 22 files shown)


niXman:
Now I'm open for discussion about code-generation for creation of tables.

niXman:
I'll add the description about `ppgen` requirements: `boost.preprocessor-1.57 or greater is requered`

niXman:
Ping?

rbock:
It is still based on the old table structure, I am afraid. I changed that end of November. It should be fairly easy to adapt, but you need to update your fork.

Also, it would be great if you could add a test minimalistic test that does something with the generated table. Since the rest of the library works without boost, it might be a good idea to check if boost is available before compiling it.

niXman:
> It is still based on the old table structure, I am afraid. I changed that end of November. It should be fairly easy to adapt, but you need to update your fork.

Ok...

> it would be great if you could add a test minimalistic test that does something with the generated table.

Ok.

> Since the rest of the library works without boost, it might be a good idea to check if boost is available before compiling it.

How do you suggest to do that?

rbock:
Checking for boost: I haven't tried it yet, but this looks promising:

http://stackoverflow.com/a/6646746/2173029

PR #27, opened 2015-02-04
enabled -Wpedantic and fixed all the warnings

TyRoXx:
I tried to use sqlpp11 and the first problem I encountered were all the warnings when including it into my project. Most of them were caused by -Wpedantic which was enabled in my project, but not in sqlpp11. I enabled -Wpedantic in sqlpp11, too, and made all instances of the newly enabled warnings disappear. Additionally, I enabled -Wconversion and corrected a truncating conversion in a piece of example code. On Ubuntu 14.04 AMD64 with GCC 4.8 sqlpp11 should be able to compile warning-free again. I tried to run the tests, but could not figure out how to make all of them run. The following was part of the output of ctest:

```
[...]
Unable to find executable: select
15/15 Test #15: select ...........................***Not Run   0.00 sec

73% tests passed, 4 tests failed out of 15

Total Test time (real) =   0.02 sec

The following tests FAILED:
	 12 - insert (Not Run)
	 13 - update (Not Run)
	 14 - remove (Not Run)
	 15 - select (Not Run)
Errors while running CTest
```

I do not think that I broke anything, but you should know that I did not run all of the tests. I did not run any of the connectors' tests either. Documentation on running the tests would be a helpful future addition to the library.


rbock:
Hi,

Thank you very much for this work. I haven't compiled with -Wconversion and -Wpedantic in years :-). I'll look into it in more detail asap, especially regarding the tests. Those should just run of course.

Regarding the changes: The majority of warnings seems to be caused by macros that have the semicolon included. Instead of removing all the semicolons from the code that uses the macros, wouldn't it be easier to just remove the semicolons from the macros?

Regards,

Roland

rbock:
The paths for ctest were incorrect in the CMakeLists.txt files. Thanks for the hint. I corrected that in the master branch.

What do you think about the semicolons? I tend to just remove them in the macros.

TyRoXx:
I did not think about that when I made the changes. The semicolons on usage seemed to be the redundant ones to me at that point in time.
I still think that macros are better when I do not have to think about semicolons when using them.

rbock:
Good point. I also looked through a few examples in boost. They seem to follow the same thought.

PR #38, opened 2015-05-18
suppress the warns about the unused vars

niXman:


include/sqlpp11/column.h
```diff
@@ -114,6 +114,7 @@ namespace sqlpp
 
 			static Context& _(const T& t, Context& context)
 			{
+				(void)t;
```
vendethiel: what about `/* t */` in the parameterlist instead?

rbock:
@niXman Thanks for looking up all those instances of unused vars. I think @vendethiel is right though, it would be better to either comment the parameter names out (or just remove them).

PR #40, opened 2015-05-19
fixes & changes to ppgen

niXman:


niXman:
Hmm... I'll fix it.
Closed.

PR #41, opened 2015-05-19
fixes & changes to ppgen

niXman:


rbock:
I could never write this PP code myself, because I cannot wrap my brains around it, but the results are really nice :-)

Thanks!

niXman:
Thank you! (got embarrassed)

PR #42, opened 2015-05-29
CMake, Travis, Coveralls

purpleKarrot:
Implemented:
* Travis CI using GCC 4.8, Clang 3.4, Boost 1.58, CMake 3.2
* Coveralls integration
* CMake overhaul

Things left to do:
* Log into https://coveralls.io/ with your Github account
* Enable rbock/sqlpp11 repository
* Change links in status badges


rbock:
Hi Daniel,

Wow, quite the cleanup :-)

Questions:
  * You removed the -std=c++11 flags. This is good because it should also work with C++14, of course. But shouldn't you add required language features then instead?
  * Why did you rename all the main functions in the tests?

Best,

Roland

purpleKarrot:
* `cxx_variadic_templates` [is required](https://github.com/purpleKarrot/sqlpp11/blob/develop/CMakeLists.txt#L37). That makes both GCC and Clang compile successfully. Other language features like `cxx_auto_type` and `cxx_constexpr` may be added. Here is the [full list of features currently known to cmake](http://www.cmake.org/cmake/help/v3.2/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#prop_gbl:CMAKE_CXX_KNOWN_FEATURES).

* Previously, each test was built to its own executable and therefore the entry point for the program (function main) was the entry point to the test. After my change, all tests are built to a single executable. This reduces compile time and the size of the build directory. The entry point for the test is a function with the signature `int(int,char**)` and the name of the test. The program entry point is [generated by cmake](http://www.cmake.org/cmake/help/v3.2/command/create_test_sourcelist.html?highlight=test_source#command:create_test_sourcelist).

rbock:
Thanks for the explanation. I added few more required language feature.

Also, I just saw that you cleaned up the CMake code for the constraint tests. Very nice :-)

PR #45, opened 2015-06-01
Do not set gcc/clang options when using msvc

jhunold:
You may want to  set extra flags with msvc...


rbock:
Thanks Jürgen, this will make life easier for whoever tries next, maybe with MSVC 2016...

PR #51, opened 2015-08-05
Remove trailing spaces.

rettichschnidi:
My IDE is configured to clean up the entire source code file when saving it. Therefore, I would like to see those purely cosmetic changes upstreamed.


rbock:
Thanks!

PR #60, opened 2015-12-28
added msvc workaround & fixed minor bugs in tests

isliser:


rbock:
Almost there, it seems (although clang/gcc is broken now)

For MSVC, for some reason, the linker is totally unhappy and there also is a problem in the assert tests.

isliser:
I'm trying to find out what's wrong

rbock:
Thanks! I'll be back in the evening.

isliser:
I suppose that AppVeyor build fails because of build\examples\example_main.cpp file. In this file methods insert, remove, select & update are declared as int(int,char * * const) whereas they are defined as int(int, char**)

rbock:
The build\tests\test_main.cpp is generated the same way by cmake. Any idea why it would work in tests, but not examples?

isliser:
It fails on all projects. You can see it https://ci.appveyor.com/project/rbock/sqlpp11/build/1.0.12/job/sql561kr9dj7vaaa

rbock:
Hmm, seems I did not wait long enough before searching for those errors. Will try to figure something out in the evening.

isliser:
Fix no_conversion_operator_if_null_not_trivial please. I cannot comprehend the idea behind it :(

rbock:
The idea is that when NULL is not interpreted as trivial and the value can be NULL, then the conversion `int i = row.alpha` is not valid.

I am not particularly fond of the tests in that folder myself. I will replace them sooner or later...

rbock:
I'll merge this pull request, try to fix the example/tests folders and then see what I can do about the conversion_operator thingy...

rbock:
Oh, wow, you fixed the tests already! Thanks!

rbock:
Well, that'll teach me: The test did not even work correctly. I'll repair that (and think of a new test, shortly).

Issue #18, opened 2014-10-01
Expression tree serializer should simplify parentheses


thomasmarsh:
Certain queries are not possible to express with sqlpp11 due to excessive parentheses that are generated during serialization. For example, the query `select(x).from(A).where(x.not_in(select(y).from(B)))` will create an extra set of parentheses around the `SELECT` statement in the `IN` clause. This breaks on sqlite3. As an example (run with `sqlite3 -echo < test.sql`):

```
DROP TABLE IF EXISTS A;
CREATE TABLE A (x INTEGER);
DROP TABLE IF EXISTS B;
CREATE TABLE B (y INTEGER);

INSERT INTO A (x) VALUES (1), (2), (3);
INSERT INTO B (y) VALUES (1), (2), (3);

-- OK: correctly returns no results
SELECT x FROM A WHERE x NOT IN (SELECT y FROM B);

-- ERROR: returns '2' and '3' - this is the format that sqlpp11 generates
SELECT x FROM A WHERE x NOT IN ((SELECT y FROM B));
```

PR #61, opened 2016-01-09
Added cmake package config file

ianda:
Now sqlpp11 can be used with

find_package(Sqlpp11)
target_link_libraries(... sqlpp11)

in your project's CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ endif()
 target_include_directories(sqlpp11 INTERFACE
   $<BUILD_INTERFACE:${DATE_INCLUDE_DIR}>
   $<BUILD_INTERFACE:${sqlpp11_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:include>
   )
 
 if (NOT MSVC)
@@ -64,10 +65,49 @@ target_compile_features(sqlpp11 INTERFACE
   )
 endif ()
 
+install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/sqlpp11"
+        DESTINATION include
+)
+
+install(TARGETS sqlpp11
+        EXPORT Sqlpp11Targets
+)
+
+include(CMakePackageConfigHelpers)
+
+write_basic_package_version_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11ConfigVersion.cmake"
+  VERSION ${Upstream_VERSION}
+  COMPATIBILITY AnyNewerVersion
+)
+
+export(EXPORT Sqlpp11Targets
+  FILE "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11Targets.cmake"
+)
+
+configure_file(cmake/Sqlpp11Config.cmake
+  "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11Config.cmake"
+  COPYONLY
+)
+
+set(ConfigPackageLocation lib/cmake/Sqlpp11)
+install(EXPORT Sqlpp11Targets
+  FILE
+    Sqlpp11Targets.cmake
+  DESTINATION
+    ${ConfigPackageLocation}
+)
+
+install(
+  FILES
+    cmake/Sqlpp11Config.cmake
+    "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11ConfigVersion.cmake"
+  DESTINATION
+    ${ConfigPackageLocation}
+)
+
+
 add_subdirectory(tests)
 add_subdirectory(test_static_asserts)
 add_subdirectory(test_constraints)
 add_subdirectory(examples)
-
-install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/sqlpp11" DESTINATION include)
-
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,7 @@ include(CMakePackageConfigHelpers)
 
 write_basic_package_version_file(
   "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11ConfigVersion.cmake"
-  VERSION ${Upstream_VERSION}
+  VERSION ${PROJECT_VERSION}
   COMPATIBILITY AnyNewerVersion
 )
 
--- a/cmake/Sqlpp11Config.cmake
+++ b/cmake/Sqlpp11Config.cmake
@@ -1,0 +1,26 @@
+# Copyright (c) 2016, Christian David
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#   Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+#
+#   Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -64,10 +65,49 @@ target_compile_features(sqlpp11 INTERFACE
...
+
+install(TARGETS sqlpp11
+        EXPORT Sqlpp11Targets
+)
+
+include(CMakePackageConfigHelpers)
+
+write_basic_package_version_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/cmake/Sqlpp11ConfigVersion.cmake"
+  VERSION ${Upstream_VERSION}
```
rbock: How/where is `Upstream_VERSION` being set? Is that taken from git tags somehow?

ianda:
The usage of ```Upstream_VERSION``` was a copy and paste error. Btw: the hole code is based on https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html

ianda:
I forgot to mention: This does not handle the date.h library. As this is a single file, header only library you may want to consider just adding it into the sqlpp11 repository. Otherwise a separate ```…Config.cmake``` file for it will be needed — which seems a bit overpowered.

rbock:
I thought about adding date.h to the sqlpp11 repository. But it seems a bit weird to me. It is the best date library I have ever seen (IMHO). I therefore expect a growing number of people to have the library installed anyway. Having a separate version as part of sqlpp11 is then bound to lead to conflicts.

ianda:
I thought about this issue. I see these options to solve this:

1) Add a findDate.cmake to sqlpp.
2) Add a cmake config file to the date library.
3) Ship sqlpp with ```date.h``` but add a cmake option to use your own ```date.h```.
4) Howard gives the library a unique name (and then uses option 1 or better 2).

I dismissed option 1. The issue is that "date" is not a generic name at all. Apparently there are a couple of libraries which include a "date.h" so the find algorithm must look into the file to verify that the right library is found. This is error prone and a lot of work.

Adding a cmake config file to date library is probably a bit too much. This would add two files just to find a single one. However I could do that quite quickly (if Howard is okay with this option). Also how would the package be called then? ```find_package(date)``` could lead to name clashes.

So the best option for me seems to be the third one. The date.h repository could be added as an git submodule or you (sorry for putting work on you!) just copy the "date.h" and update it if needed. By default this "internal" date.h is used and installed with sqlpp. If the user wants to change that, I’ll provide a cmake option which allows him to the define the location of his own date.h. Then besides using ```find_package(Sqlpp11)``` the user has to ensure that the right version of date.h can be found.

Option 4 for a single file could be too much again. Also it would make sense if the timezone functions are used. However, Sqlpp does not use this feature anyway.

To prevent conflicts of different date.h versions, we need to convince Howard to use versions. Otherwise I see no reasonable way of auto checking if there are any conflicts.

Also; the connectors install binary code. So at compile time, the version of date.h gets fixed. Without a version we could compare to, conflicts are still possible.

What is your opinion on this?

rbock:
The easy parts first:
- Option 1: I agree, the name `date.h` is rather generic.
- Option 2: Right, and I know that Howard dismissed the idea of having a CMakeLists.txt in his repo just a few weeks ago.
- Option 4: Renaming date.h is probably not going to happen any time soon.

About option 3:
I would like to avoid a submodule as it complicates the git part of the library. So I _could_  add a local copy of date.h and then install it, too. I need to think about that...

But: I would certainly not make that version of the library the default. If somebody wants to use this, they should specify their intent explicitly.

Here is my current preference (not final, I just woke up):

In the `CMakeLists.txt`: Add an option to specify the path to the date library and then check if we can compile a mini date-test program using
  1. the specified `PATH_TO_DATE_LIBRARY`, if a path is given
  1. the local copy, if `PATH_TO_DATE_LIBRARY` is specified as `SQLP11_LOCAL`
  1. the default include paths (maybe it is just installed in `/usr/include`?)
  1. (sqlpp11-root)/../date as `PATH_TO_DATE_LIBRARY`
The first option to compile successfully wins. If nothing compiles, give a decent error message.

In the cmake-package it would be quite similar, except that the `../date/` probably needs to be omitted.

snikulov:
> I would like to avoid a submodule as it complicates the git part of the library

Submodule can give you some dependency version control.

ianda:
Looking for date.h in different folders and then compiling a test program is a variant of adding a ```findDate.cmake``` – but a complex one. Compiling test programs during the configuration step does not sound like a good solution to me. Also I would not like to implement that.

I did a quick search on the internet and could not find any Linux package which includes the date library (directly or indirectly). So maybe it is useless to implement a find routine for a pre-installed date.h anyway (e.g. in ```/usr/include```). If the library becomes used by more developers we can still adopt to that (then there is also a high chance someone else will solve the issue).

If you want, I can implement option 3 in a way that it will fail by default. Only if the option points to folder which includes a date.h it will succeed. Of course this makes the compile process more annoying but prevents from using the internal date.h by default.

Btw: You said Howard rejected any CMakeLists.txt in his repository, do you mean [Issue #18](https://github.com/HowardHinnant/date/issues/18)? If I add cmake config files I might be able to convince him that it has a benefit. But that still does not solve the ambiguous name.

rbock:
On 2016-02-16 17:56, ianda wrote:
> Looking for date.h in different folders and then compiling a test
> program is a variant of adding a |findDate.cmake| – but a complex one.
> Compiling test programs during the configuration step does not sound
> like a good solution to me. Also I would not like to implement that.

I'll look into that sometime soon. I thought there is some test_compile
function in cmake.

> Btw: You said Howard rejected any CMakeLists.txt in his repository, do
> you mean Issue #18 <https://github.com/HowardHinnant/date/issues/18>? If
> I add cmake config files I might be able to convince him that it has a
> benefit. But that still does not solve the ambiguous name.

That seems to be the underlying problem...

ianda:
There is a cmake function to test if something compiles : [```try_compile```](https://cmake.org/cmake/help/v3.5/command/try_compile.html).

I would just prefer to look into the file and check for a line ```Copyright (c) ⟨4 digits⟩ Howard Hinnant``` and maybe ```namespace date```. Then then there should be a low chance that a wrong file was found. This method should be way waster than compiling test programs (I hate these autotools configure scripts which need forever to compile a dozen tests).

rbock:
On 2016-02-16 19:12, ianda wrote:
> I would just prefer to look into the file and check for a line
> |Copyright (c) ⟨4 digits⟩ Howard Hinnant| and maybe |namespace date|.

OK

> (I hate these autotools configure scripts which need forever to compile a dozen tests).

:-)

rbock:
Sorry it took so long. I merged your PR.

I like the idea of testing for Howard Hinnant in the date file.

Erroneous1:
In the Arch Linux User Repository (AUR) the sqlpp11 package is putting date.h inside of the sqlpp11 include directory. In my brewpipp project I'm working on I use the following to use that:

`if(EXISTS "${sqlpp11_INCLUDE_DIR}/sqlpp11/date.h")
	file(MAKE_DIRECTORY "${PROJECT_SOURCE_DIR}/date")
	file(WRITE "${PROJECT_SOURCE_DIR}/date/date.h" "#include <sqlpp11/date.h>\n")
	set(sqlpp11_INCLUDE_DIRS ${sqlpp11_INCLUDE_DIRS} "${PROJECT_SOURCE_DIR}/date")
	message("date.h found in ${sqlpp11_INCLUDE_DIR}/sqlpp11/date.h")
endif()`

PR #83, opened 2016-05-01
Added FindHinnatDate.cmake to prevent usage of preset paths

ianda:
Howard Hinnant's date and time library is found using a cmake Find-module.

Commit 377fbc8958b26cb2c0151b3068bce07dff3811ec was reverted to ensure that the
exported target has the correct properties.


rbock:
Hi,

Thanks for writing the Date lookup! The CI builds failed, though. I assume this is because I do not install the date library?

I guess there should be a few optional paths in the script like 
"${CMAKE_CURRENT_SOURCE_DIR}/date" 
"${CMAKE_CURRENT_SOURCE_DIR}/../date"

Also, can I set a variable to the correct path? It would be great if the find-module said so. Or the main CMakeLists.txt should say so if date is not found.

ianda:
If the date library is not installed in a default location, you should call cmake with the path to look in, e.g.:

    cmake -DCMAKE_PREFIX_PATH="path/to/date" ../path/to/sqlpp11-source

Alternatively you could set the new ```HinnantDate_ROOT_DIR```. The find module notifies the user about the availability of this variable now.

If you want to ship the date library with sqlpp11, you should set ```HinnantDate_ROOD_DIR``` before calling ```find_package```. The find module looks in default paths already (see manual for [find_file](https://cmake.org/cmake/help/latest/command/find_file.html)). ```${CMAKE_CURRENT_SOURCE_DIR}/../date``` is not part of these default paths. For good reason, I think.

rbock:
OK, agreed.

I do not intend to ship date as part of sqlpp11. The user should set the path (if it is not installed in the default location).

Thanks for the effort!

ianda:
As you can see I really tried to get AppVeyor working. Now I am at a point where I have no clue what is going wrong.

Also worth to mention: Unfortunately ```date.h``` does not offer any versioning I can use (e.g. a version number inside the file). So this cannot be checked.

rbock:
According to the build log, the path to date should be

C:\projects\date

I guess you have to escape the backslashes? Or maybe you can use forward slashes with CMake?

rbock:
Cool, thanks for the effort!!

PR #84, opened 2016-05-02
added common SQL data types like binary and bit

Erroneous1:


include/sqlpp11/data_types/time_point/data_type.h
```diff
@@ -40,5 +40,9 @@ namespace sqlpp
     template <typename T>
     using _is_valid_operand = is_day_or_time_point_t<T>;
   };
+
+  using datetime = time_point;
+  using timestamp = time_point;
+  using time = time_point;
```
rbock: datetime and timestamp are certainly OK, as long as everybody agrees on the timezone.

But are you sure about time? Doesn't that represent time of day?

Erroneous1:
DATETIME is localized to session/server time in text but in ODBC when using bound results it comes out UTC. It seems that this would be sqlpp connector dependent. MS has an article on how times are converted [https://msdn.microsoft.com/en-us/library/bb677170.aspx](here) and specifies (if you assume they typoed 23 when they meant 22 for SQL_C_SS_TIMESTAMPOFFSET) that the time is converted from client to UTC before being put into a SQL_TYPE_TIMESTAMP (which is what I use in the ODBC connector). If it isn't already we should specify that UTC is to be assumed when working with sqlpp::chrono::microsecond_point.

I could see an argument for needing to add a time-only data type or specialize datetime to also have an offset, but at least with some connectors this would work if you ignore the date held in the results from a column known to be a time only column.

For time only columns it returns the correct time but with today's date using ODBC connector. If you try my ODBC connector's test code [https://github.com/Erroneous1/sqlpp11-connector-odbc/blob/master/tests/ODBCTest.cpp](tests/ODBCTest.cpp) you can test.
Quick breakdown of what it does:
Create the table
`CREATE TABLE tab_bar (
                        delta date DEFAULT NULL,
                        epsilon datetime DEFAULT NULL,
                        zeta timestamp,
                        eta time DEFAULT NULL
);`
insert the following:
`INSERT  INTO tab_bar (delta,epsilon,zeta,eta) VALUES(DATE '1970-01-02',TIMESTAMP '1970-01-02 04:48:43.123456000',TIMESTAMP '1970-01-02 04:48:43.123456000',TIMESTAMP '1970-01-02 04:48:43.123456000')`
Then select the values and prints them. In a quick modification, I changed the date to 1970-01-02T04:48:43.123456000 and had it SET time_zone 'UTC', print the selects, then SET time_zone 'America/New_York' and had the following results:

UTC
> delta:  1970-01-02, epsilon:    1970-01-02T04:48:43.000000, zeta:       1970-01-02T09:48:43.000000, eta:        2016-05-04T04:48:43.000000

America/New_York
> delta:  1970-01-02, epsilon:    1970-01-02T04:48:43.000000, zeta:       1970-01-02T04:48:43.000000, eta:        2016-05-04T04:48:43.000000

This was on unixodbc connecting to MySQL running on linux with zoneinfo loaded. Without zoneinfo tables it errored out when I try to set the time_zone.

Erroneous1:
Somehow I missed that the zeta TIMESTAMP column in the above test did in fact change the time between the 2 timezones. Looks like going from SQL_SS_TIMESTAMPOFFSET to SQL_C_TIMESTAMP converts to the client timezone. The solution (for ODBC) would be to use SQL_C_SS_TIMESTAMPOFFSET as the destination data type but it isn't available for unixODBC.

@rbock Should I submit a new PR without timestamp or would it be better to just have the connectors create their own datatype using lines depending on what is supported properly?

rbock:
I believe that the second option (having connector-specific types) is more appropriate.

Erroneous1:
After doing a couple of tests in MariaDB it seems that inserting TIMESTAMP '1970-01-01 00:00:00' does not result in FROM_UNIXTIME(0) (unless you are at offset 00:00), so it seems that specifying TIMSTAMP for time_point doesn't work as (I assume) intended for all DBMS. Also it seems MariaDB won't properly convert an iso 8601 format timestamp with offset and simply interprets it as if it was from @@session.time_zone. This means on DATETIME and TIMESTAMP data types that NOW() (local) could be different from std::chrono::system_clock::now() (UTC). To make matters worse, there isn't any way to have all DBMS interpret a date and time as UTC. If I wanted to do a DBMS specific way to do this for the ODBC connector I'd need access to the db object to know which function to use, is this possible to do when serializing? Also, for prepared statements, I'd need to apply a function to the parameter and I'm not sure that is currently possible.

Other than date and time functions, DATETIME should be the column type used by sqlpp11 and the database design should assume that it is un UTC. If you need to use NOW() then you should resort to std::chrono::system_clock::now() or convert it to UTC.

On another note, I've been working on adding a time with no date value to the ODBC connector and I think it probably be better to add a time-only data type to sqlpp11 instead of making it connector specific. I'm planning on doing the following:
- use sqlpp::chrono::microsecond_point as the value type
- call _bind_time_result, _post_bind_time_result, and _bind_time_parameter.
- I'm thinking of calling it either time_only or just time.

Also, would it be beneficial to add a uint64_t data type? That way you have less issues with comparison between signed and unsigned numbers.

rbock:
How about calling it `time_of_day`?

And yes, a uint64_t would certainly be good.

rbock:
I finally merged the other two pull requests. I suggest you remove the timepoint before I merge this one?

rbock:
Thanks! Merged to develop.

Once I fix the compilation error on MSVC, I'll create a new release.

Best,

Roland

PR #90, opened 2016-05-19
feature/time_of_day

Erroneous1:
This uses std::chrono::microseconds as the value type instead of sqlpp::chrono::microsecond_point. To convert a microsecond_point to microseconds, a helper function was added to sqlpp::chrono called time_of_day. Added tests to DateTime.cpp and a column definition to Sample.h.


rbock:
Merged into develop.

PR #91, opened 2016-05-19
Feature/unsigned integral

Erroneous1:
Added unsigned integral which uses uint64_t


rbock:
Thanks for adding those types. I'll do some review on the weekend.

PR #92, opened 2016-05-20
feature/unsigned_integral

Erroneous1:
reset to master, added unsigned_integral, removed time_of_day
Ran ctest to make sure everything works.

### DIFF
--- a/include/sqlpp11/data_types/unsigned_integral/expression_operators.h
+++ b/include/sqlpp11/data_types/unsigned_integral/expression_operators.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013-2015, Roland Bock
+ * Copyright (c) 2013-2016, Roland Bock, Aaron Bishop
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification,
--- a/include/sqlpp11/data_types/unsigned_integral/expression_operators.h
+++ b/include/sqlpp11/data_types/unsigned_integral/expression_operators.h
@@ -1,0 +1,108 @@
+/*
+ * Copyright (c) 2013-2015, Roland Bock
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ *   Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ *   Redistributions in binary form must reproduce the above copyright notice, this
+ *   list of conditions and the following disclaimer in the documentation and/or
+ *   other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
+ * ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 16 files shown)


include/sqlpp11/data_types/unsigned_integral/expression_operators.h
```diff
@@ -0,0 +1,108 @@
...
+  {
+    using check = consistent_t;
+    using type = unary_plus_t<unsigned_integral, wrap_operand_t<T>>;
+  };
+
+  template <typename T, typename Defer>
+  struct return_type_unary_minus<T, Defer, unary_operand_check_t<T, is_unsigned_integral_t>>
+  {
+    using check = consistent_t;
+    using type = unary_minus_t<unsigned_integral, wrap_operand_t<T>>;
```
rbock: I would assume that this should return a signed integral.

include/sqlpp11/data_types/unsigned_integral/operand.h
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright (c) 2013-2015, Roland Bock
```
rbock: Here and in the other new files, switch to 2016 and add your name.

include/sqlpp11/data_types/unsigned_integral/expression_operators.h
```diff
@@ -0,0 +1,108 @@
...
+  {
+    using check = consistent_t;
+    using type = plus_t<wrap_operand_t<L>, value_type_of<wrap_operand_t<R>>, wrap_operand_t<R>>;
+  };
+
+  template <typename L, typename R>
+  struct return_type_minus<L, R, binary_operand_check_t<L, is_unsigned_integral_t, R, is_numeric_t>>
+  {
+    using check = consistent_t;
+    using type = minus_t<wrap_operand_t<L>, value_type_of<wrap_operand_t<R>>, wrap_operand_t<R>>;
```
rbock: In SQL, if you subtract two unsigned values, say 3 - 7, the result is signed.

tests/SelectType.cpp
```diff
@@ -121,6 +122,23 @@ int SelectType(int, char* [])
...
+    static_assert(sqlpp::is_selectable_t<T>::value, "type requirement");
+    static_assert(not sqlpp::require_insert_t<T>::value, "type requirement");
+    static_assert(not sqlpp::is_boolean_t<T>::value, "type requirement");
+    static_assert(not sqlpp::is_text_t<T>::value, "type requirement");
+    static_assert(not sqlpp::is_alias_t<T>::value, "type requirement");
+    static_assert(not sqlpp::is_table_t<T>::value, "type requirement");
+    static_assert(sqlpp::is_regular<T>::value, "type requirement");
+  }
+
   // Test a floating point table column
```
rbock: Given the comments above, it might make sense to add a few more tests here.

include/sqlpp11/data_types/unsigned_integral/expression_operators.h
Erroneous1: Would I just change the wrap_operand_t to wrap_operand_t<integral> for L and R? Is there any kind of conversion I need to do?

rbock: You replace
```
minus_t<wrap_operand_t<L>, value_type_of<wrap_operand_t<R>>, wrap_operand_t<R>>;
```
with
```
minus_t<wrap_operand_t<L>, integral_t, wrap_operand_t<R>>;
```

rbock: Oh, not good yet. If `R` is float, then the result is float, too. Maybe you need to make unsigned_integral a special case that returns integral. Otherwise it returns the value_type of R.

rbock: I just realized that this also affects operators for`integral_t`.

Maybe it would be best to have a central meta function to calculate the result type of arithmetic operations for all available types.

Would you like to do that? Otherwise, I can do it, too, of course, just let me know.

Erroneous1: Looks like it also changes to signed for modulo, add, multiply, and divide. Similar for integrals and floats. I think I can do it if you'd like.

rbock: Why would it change to signed when you add/multiply/divide two unsigned?

Erroneous1: -3 + 1 = -2
-1 * 5 = -5
-2 / 1 = -2
sqlite3 behaves this way, idk about what ANSI SQL says (or it unsigned is part of ANSI SQL). Also, in MySQL (specifically MariaDB, but MySQL is probably the same) if you subtract an unsigned column from an unsigned column it throws an error if the result would be negative and the . This is different than sqlite3 which just returns a signed version.

I did a commit to fix integrals and unsigned integrals so that any operation between a signed and unsigned will yield a signed (except modulus which I left to same type), and added tests for unsigned to make sure that any type and any op with an unsigned yields the other type and any time you subtract or unary minus an unsigned then it is signed. I added a type trait for is_numeric_not_unsigned_t and specialized some of the expressions for integral when working with unsigned to return signed.

rbock: Nice! I'll look more detail ASAP (latest next weekend) and merge this into develop first.

rbock:
Merged into develop (the build problem on windows is due to another change I did).

PR #100, opened 2016-06-09
Readme for date lib

digitalist:
just some readme fixes, so people don't get frustrated if cmake . does not work


rbock:
Thanks :-)

PR #102, opened 2016-06-12
ddl2cpp tests

digitalist:
This PR adds ddl2cpp python script tests.
- cmake-time check for python / pyparsing
- ddl2cpp good and bad samples
- should work for offline tests (it does so for me :-)
- works on travis
-  ddl2cpp tests on osx are fake/disabled for now

that's it. Now I know a lot more about Travis.
and a lot more about cwd() on linux/windows/osx. Wish I never googled this, but it's ok.


digitalist:
@rbock by the way, do you have any plans on implementing travis build caching?

rbock:
Nice, and you're certainly not stupid :-)

PR #118, opened 2016-09-10
fixed missing include for return_type_not_t

Erroneous1:
I'll need this before I can publish a 0.42 compatible version of sqlpp11-connector-odbc.


rbock:
Thanks, I guess I should check all include files if they have complete include lists.

Opening an issue now...

PR #125, opened 2016-10-20
Find Hinnant date library in date/ subfolder

Marvin182:
Horwards date library has grown to a few headers and I think it is cleaner to have them in a subfolder of the standard include path.
This is in line with my homebrew formula for date/sqlpp11: https://github.com/Marvin182/homebrew-zapfhahn/tree/master/Formula

### DIFF
--- a/cmake/Modules/FindHinnantDate.cmake
+++ b/cmake/Modules/FindHinnantDate.cmake
@@ -53,7 +53,7 @@ set(HinnantDate_NOT_FOUND_MESSAGE "Could NOT find HinnantDate.
 Maybe you need to adjust the search paths or HinnantDate_ROOT_DIR.")
 
 find_file(HinnantDate_INCLUDE_FILE
-    date.h
+    date.h date/date.h
     HINTS ${HinnantDate_ROOT_DIR}
 )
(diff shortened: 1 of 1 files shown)


rbock:
Marvin,

Thanks! I like the name "homebrew-zapfhahn" :-)

I understand from your README that your formulas are not well enough tested to be integrated into main tap. Would you say it makes sense to point others to them, though? If so, please create another pull request for the README to add a link. I'd say next to the Linux install text.

Marvin182:
Hi :)

Thoughts on that:

1) I created a pull requests with an install section, describing both install in general and with homebrew.

2) My formula only contains options for the sqlite3 and mysql connector. Authors of other connectors are welcome to add install instructions and dependencies for their connectors.

3) The formula has an --HEAD option to use the latest version of the master branch, but for the stable version we have to update the formula after each release. I might be the wrong person to keep track of that, but you can create pull requests on to homebrew-zapfhahn for it.

4) I once wanted to push a formula to one of the official taps and it was a little disasters as they kept coming back with comments about little style changes I should do. Those style requirements weren't documented. When send be back and forth between two taps I decided to make my own tap. Everybody can use them, they are just harder to find.

PR #132, opened 2017-01-07
Allow nested namespaces on the command line

mlimber:
Allow the generator to handle nested namespaces like:
`ddl2cpp table.ddl table My::Nested::Namespaces`


rbock:
Thanks!

PR #141, opened 2017-02-21
Fix variadic template pack expansion failures on VS2017 RC

fpark12:
This pull request includes 2 workarounds to fix compile errors that occur on VS2017 RC but not VS2015. With this patch sqlpp11 compiles successfully with VS2017 RC.

### DIFF
--- a/include/sqlpp11/having.h
+++ b/include/sqlpp11/having.h
@@ -237,7 +237,7 @@ namespace sqlpp
       //	  template <typename... T>
       //	  using _check = logic::all_t<is_expression_t<T>::value...>;
       template <typename... T>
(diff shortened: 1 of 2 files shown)


rbock:
Hi,

This is trying to please an obviously seriously broken compiler. This is breaking other compilers.
Of course, you _could_ use preprocessor to choose one way or the other, but I have no way of maintaining that.

The only reasonable thing to do here is file a bug against the compiler and get Microsoft to fix it.

Best,

Roland

fpark12:
I agree, and I hope Microsoft will fix it in the future. I'll submit a bug report soon.

Frank

Issue #142, opened 2017-02-28
Some enhancement suggestions for result_t::iterator


theodelrieu:
Hi, awesome library!

It would be really cool if that iterator could be a fully standard conformant one.
For example, I just had an error with MSVC looking for `std::iterator_traits<result_t::iterator>::iterator_category`, while doing something of the sort:

```cpp
// 'auto const' results in a compile error (at the transform call), don't know why
auto rows = db(select(all_of(user)).from(user).unconditionally());
std::vector<cpp_user> vec;

// this breaks on MSVC 2015
std::transform(rows.begin(), rows.end(), std::back_inserter(vec), convert_to_user);
```

1. The `transform` call can be replaced by a for-loop, so that's not a really blocking issue.
2. I have no idea why I cannot store the result of `db(...)` as `auto const` ?

I don't fully remember what's needed for an iterator to be fully conformant, but providing the specialization of `std::iterator_traits` would be really nice.

PS:
I might open another issue for this one, but isn't it possible to have a `size` method on the row object?
My use-case is to call `reserve` on a vector, and right now I have to do a `select(count(...))` to have that info (`std::distance` doesn't work because of `iterator_category` too)

Thanks in advance!

rbock:
Dear Theo,

Thanks for the detailed issue.

Let's start with the `size`. I think there is no portable way to obtain the size for the result set. Some databases just don't offer this information. However, it would be possible to offer it for those databases that have the information, of course.

About the non-constness of the result set: This is similar to an `istream`: It changes internal state while you read it. Depending on the database (and/or the type of connection) you can only forward-iterate the result set, e.g. rows are read on demand from the server and discarded after being read.

Regarding the iterator conforming the requirements, I guess there isn't much missing.

It would be awesome if you could help with this. I think it would be best to start with the iterator. Even if you can't fix the iterator code itself (I know the code isn't easy), you could provide one or more test cases that should work?

Size would be next. I think it would make sense to open a second issue for that.

Cheers,

Roland

theodelrieu:
I see, I'll try to fix the iterator this week then!

rbock:
Great! Let me know if you need support!

rbock:
The iterator issue is fixed, I think. Thanks for your support!

If you want to tackle size(), please create a new pull request.

PR #143, opened 2017-03-01
Enhancements to result_t::iterator

theodelrieu:
Hi Roland,

The first commit deals with the `iterator_traits` requirements we talked about in #142 .
About the second, I remember you said the iterator mutates the underlying sequence, therefore there could be no const overloads of `begin` and `end`.

I do think however, that the underlying mutation is not a part of the logical interface of the result object, but merely an implementation detail (part of the physical interface).

Furthermore, the following code already fails to compile:

```cpp
auto rows = db(select(all_of(t)).from(t).unconditionally());
for (auto& row : rows)
{
    // row is const here
    row.alpha = 4;
}
```

It would make sense (in my opinion), to be able to do `const auto rows = /* ... */` as well.

I'm looking forward to your inputs!

### DIFF
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -27,9 +27,17 @@
 #ifndef SQLPP_RESULT_H
 #define SQLPP_RESULT_H
 
-// FIXME: include for move?
+#include <iterator>
+#include <utility>
+
 namespace sqlpp
 {
+  template <typename>
+  struct iterator_category
+  {
+    using type = std::input_iterator_tag;
+  };
+
   template <typename DbResult, typename ResultRow>
   class result_t
   {
@@ -60,16 +68,22 @@ namespace sqlpp
     class iterator
     {
     public:
+      using iterator_category = typename iterator_category<DbResult>::type;
+      using value_type = result_row_t;
+      using pointer = const result_row_t*;
+      using reference = const result_row_t&;
+      using difference_type = std::ptrdiff_t;
+
       iterator(db_result_t& result, result_row_t& result_row) : _result(result), _result_row(result_row)
       {
       }
 
-      const result_row_t& operator*() const
+      reference operator*() const
       {
         return _result_row;
       }
 
-      const result_row_t* operator->() const
+      pointer operator->() const
       {
         return &_result_row;
       }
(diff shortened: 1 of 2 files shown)


rbock:
Hi Theo,

Thanks for the quick work. I'll have to study it in more detail. Some remarks:

```
for (auto& row : rows) {}
```

This should compile and IMO it should be the same as 

```
for (const auto& row : rows) {}
```

However, neither of the following should compile:

```
for (auto row : rows) {}
for (const auto row : rows) {}
```

rows hold their text data in `char*` that are managed by the result set in the connector library. Copying this would be weird, I think. When you continue to iterate through the results, the copy would change. That would be rather strange behavior.

I do not agree that the changes inside the result set are just an implementation detail. They have consequences for the interface user.

For instance, when you advance the iterator, the begin function will return different things. Also, the `empty()` function will eventually return true once you reached the end.

Best,

Roland

theodelrieu:
Ah, I was not clear enough in my example, sorry about that.

`auto&` englobes `auto const&`, so my point was rather about the assignment in the for loop, and that the user has no way of modifying the rows, (except implicitely with the iterators, from what I've understood so far)

I fully agree that copies of the underlying objects should be forbidden.

In the second commit, I simply added const overloads of begin/end, which calls the same method than their non-const counterparts (the only difference being the `const_cast`), and thus return the same thing.

But I get your point, at first I saw `rows` as a container of row, hence my surprise.
It might confuses users which relies on the `rows` const-ness (i.e. not being modified at all), although I don't know if that is a real use-case.

I will think about this, thanks for your answer!

rbock:
Hi Theo,

What's the status? I am basically just asking to make sure that we're not both waiting on the other :-)

Best,

Roland

theodelrieu:
Hi! I didn't have that much time this week, I think I'll remove my second commit, since it's not that big of a deal for me.

I'll do that tonight normally.

rbock:
Looking forward to it :-)

include/sqlpp11/detail/get_nested_type_or.h
```diff
@@ -0,0 +1,59 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef SQLPP_DETAIL_GET_NESTED_TYPE_OR_H
+#define SQLPP_DETAIL_GET_NESTED_TYPE_OR_H
+
+#include <type_traits>
+
+#ifdef SQLPP_DETAIL_GET_NESTED_OR
```
rbock: Instead of having a macro here, we could define an iterator_category trait that defaults to `input_iterator` and can be specialized by connections.

theodelrieu:
Indeed, that's far better.

rbock:
Awesome :-)

Issue #152, opened 2017-03-19
put quotes around table and column names to allow reserved words in names


sergeken:
Putting quotes around table and column names in the SQL statements would allow to use reserved words.
For example "insert into posting set order=123" as order is a reserved word.

rbock:
Please see also  #138 opened by @cszawisza

Maybe you could join forces :-)

sergeken:
Why not.

Erroneous1:
Can you just make the column's name "\"order\"" when you create the sqlpp templates for it?

sergeken:
This works very well.  With all our discussion, I'm much more in favor of this approach.  Ideally at some time we could enhance the generator to have a smart escape-generating code with support for various RDBMS's and only for objects with reserved words.
I would for now rather update the documentation describe the manual workaround for objects that uses reserved names.  And propose to reject my pull request.

rbock:
Sounds like a very good plan to me. Would you like to add the respective documentation?

sergeken:
Hi Roland,

Here is an attempt.  Regards,
Serge
[Tables.md.txt](https://github.com/rbock/sqlpp11/files/923730/Tables.md.txt)

rbock:
Thanks! 

Added to the wiki :-)

rbock:
Merged into #199

Issue #138, opened 2017-02-02
Taking table name into " " in connector library


cszawisza:
Hi!
I'm using postgresql connector in my project, and recently I found a BUG. Connector lib don't put table names into "" and if this name is a sql keyword, postgres qill generate an error simmilar to:

>  syntax error at or near "".""
> LINE 1: ...entity FROM user INNER JOIN auth_identity ON (user.uid=auth_...

So I assume that I need to override 
`template <typename T>  std::ostream& operator<<(T t) { return _os << t;  }`
for table and field types, but don't know exactly how, can someone help me out?

rbock:
For a minute I was tempted to say: "Just don't use SQL keywords as table names." For instance, I prefix every table name with `tab_` to avoid such problems. But then again, you could have the same problem with column names and aliases, etc. In fact, the aliases for functions have a `_` suffix to prevent this problem.

In order to get a better understanding of where the serialization code is located, take a look here, for instance:

include/sqlpp11/table.h
include/sqlpp11/column.h

The serializer_t at the bottom of those files is what you want to look at.

However, if you want to fix the problem for all user-defined names and aliases, a better place might be this:

include/sqlpp11/char_sequence.h

We could insert leading and trailing ticks in the `char_ptr`  function. We could also `static_assert` that there are no ticks in names.

cszawisza:
Hi!
 I found couple of ways to "fix" this
1. `static constexpr const char _literal[] = R"("user")";` // insted of "user"
2. `using _name_t = make_quoted_char_sequence<sizeof(_literal), _literal>;` which is a modified version of make_char_sequence<...>

For now I'll use method 1. and hope that everything will work :)

rbock:
Hi,

Those are both nice ideas! Did you adjust your code generator for method 1. ?

Cheers,

Roland

cszawisza:
Not yet, but definitely I will.

rbock:
See also https://github.com/rbock/sqlpp11/wiki/Tables#names-with-reserved-sql-keywords (text by @sergeken)

rbock:
Merged into #199

PR #153, opened 2017-03-19
Fix for Issue 152

sergeken:
This is a working version with working tests.
The C++ syntax is no the nicest nor more optimal.  Using raw strings would make the test cases more readable. 
With regards to the verbatim, these are as the name suggested unaltered.

### DIFF
--- a/include/sqlpp11/alias.h
+++ b/include/sqlpp11/alias.h
@@ -66,7 +66,7 @@ namespace sqlpp
     {
       serialize_operand(t._expression, context);
       context << " AS ";
-      context << name_of<T>::char_ptr();
+      context << '"' << name_of<T>::char_ptr() << '"';
       return context;
     }
   };
--- a/include/sqlpp11/column.h
+++ b/include/sqlpp11/column.h
@@ -125,7 +125,7 @@ namespace sqlpp
 
     static Context& _(const T&, Context& context)
     {
-      context << name_of<typename T::_table>::char_ptr() << '.' << name_of<T>::char_ptr();
+      context << '"' << name_of<typename T::_table>::char_ptr() << '"' << '.' << '"' << name_of<T>::char_ptr() << '"';
       return context;
     }
   };
--- a/include/sqlpp11/cte.h
+++ b/include/sqlpp11/cte.h
@@ -235,7 +235,7 @@ namespace sqlpp
 
     static Context& _(const T& t, Context& context)
     {
-      context << name_of<T>::char_ptr() << " AS (";
+      context << '"' << name_of<T>::char_ptr() << '"' << " AS (";
       serialize(t._statement, context);
       context << ")";
       return context;
@@ -279,7 +279,7 @@ namespace sqlpp
 
     static Context& _(const T&, Context& context)
     {
-      context << name_of<T>::char_ptr();
+      context << '"' << name_of<T>::char_ptr() << '"';
       return context;
     }
   };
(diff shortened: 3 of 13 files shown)


niXman:
Great!

Thank you!

rbock:
Looks good at first glance. Will look into it in more detail soon.

Question: Do all databases accept double quotes? From the top of my had: I believe that mysql uses forward ticks for quoting names.

sergeken:
Most standard conformant databases do use double quotes.  An important side note is that when using quotes, some databases become case sensitive.  This is the case for postgresql if I'm not mistaken.  With regards to MySQL I found a stack overflow article stating that MySQL in ANSI SQL mode does accept double quotes.

rbock:
Thanks for looking it up!

Maybe it would be good to get the quote-character from the `context`. Preferably in a way that standard conforming connectors do not need to do anything. Something like:

```
template <typename Context>
constexpr auto get_name_quote_character(const Context&) -> char
{
   return '"';
}
```

For MySQL I would then create a specialization that takes the character from the context. Most of the time it would be ``` ` ```, but the user could set it to `"`, too.

toby-allsopp:
Sorry to butt in, but you might like to consider that Microsoft's SQL
Server prefers to use square brackets to quote names (although I think
double quotes also work), so it might be better to have a function that
takes a name and returns a quoted version of it.

The case sensitivity thing is going to be an issue - you might need to have
an option to uppercase the name when quoting it.

rbock:
Not at all. Thanks for the information!

So basically, we need to `context` to have 
  * a function to escape quote names
  * an option to turn quoting on/off

And preferably it should be written in a way that does not break existing connector libraries.

Or we follow the idea in #138 and change the code generator. That seems much less intrusive and easier to control. However, it would not help with `alias`.

sergeken:
I've been thinking over it also.  As this is a "border" case I do not consider it an urgent issue and needs proper thinking through.  Using reserved names does bring issues with plain SQL anyway and hence to be considered a poor practice.  Moreover, it pollutes the generated SQL, unless we make it smart and finally the case sensitiveness is a real problem that has bitten me in the past when writing manual SQL statements.  If on top of that we cannot rely on standard conformance we have to make the solution very flexible.  Not sure what is the best way.  Maybe a flag to the code generator is the best approach like "-escape=ansi" or "-escape=mysql", etc. to activate RDBMS conformant escaping as usually people know the RDBMS from which they use the DDL scripts.

Erroneous1:
Using double quotes on everything can [produce issues on some RDBMS ](http://stackoverflow.com/questions/2901453/sql-standard-to-escape-column-names#comment70616858_2901502)where it will make it a string if a column name isn't available, when you might want it to actually error out because the column doesn't exist. Looking at the patch I think that might only happen with simple columns since the RDBMS should be able to tell `"valid_table"."invalid_column"` is referring to `valid_table.invalid_column` and aliases are going to be defined in the statement itself.

A simpler solution might be to put your own quotes/backticks/square brackets around columns and tables on the columns and tables when you define them. This way you wouldn't have any unexpected behavior and if you have a column or table using a reserved word then you can escape it in a fashion befitting your RDBMS. The obvious issue is that it makes it more work and you'd need different headers for different RDBMS (unless you use `"`). We could have the table generators automatically add them as well.

If we do allow for different escape types, it might be best to use 1 character for before and 1 character for after and just always do them. ANSI and MySQL would use `"` and `` ` `` for both, while MS SQL would use `[` and `]`. Then you don't need context specific function.

rbock:
Closing this pull request, see also discussion in #152.

Thanks for the input everyone!

Cheers,

Roland

PR #171, opened 2017-06-01
Add an option to set a transaction isolation level in start_transaction

volka:
This code lets a user select one of the four standardized transaction isolation levels when opening a transaction. Actual support dependes on the connectors, e.g. for SQLite only "serializable" and "read uncommitted" are meaningful.

I tried to implement the start_transaction() call in a way that should be backwards compatible with connectors not supporting the isolation_level parameter, but that makes for some possible confusion with the overloads of start_transaction.

For implementations see pull requests for the sqlite3 and postgresql connectors.

### DIFF
--- a/tests/MockDb.h
+++ b/tests/MockDb.h
@@ -28,6 +28,7 @@
 
 #include <iostream>
 #include <sqlpp11/connection.h>
+#include <sqlpp11/transaction.h>
 #include <sqlpp11/data_types/no_value.h>
 #include <sqlpp11/schema.h>
 #include <sqlpp11/serialize.h>
@@ -244,6 +245,22 @@ struct MockDbT : public sqlpp::connection
   {
     return {name};
   }
+
+  void start_transaction(sqlpp::isolation_level level)
+  {
+    _current_isolation_level = level;
+  }
+
+  void rollback_transaction(bool)
+  {}
+
+  void commit_transaction()
+  {}
+
+  void report_rollback_failure(std::string)
+  {}
+
+  sqlpp::isolation_level _current_isolation_level;
 };
 
 using MockDb = MockDbT<false>;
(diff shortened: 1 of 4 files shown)


volka:
Strange, Visual Studio does not seem to like "not" (though it supposedly should work with the /Za compiler flag). It is used quite often in the code ...

rbock:
Hi Volker,

You need to add 

```
#include <ciso646>
```

Citing cppreference (http://en.cppreference.com/w/cpp/header/ciso646):

>  It may still be required in order to have the alternative operator representations in old or non-conformant compilers.

Best,

Roland

rbock:
This is a nice addition.

I am not a big fan of default arguments. Maybe we could have two overloads (with/without isolation level)?

volka:
Yes, I thought that might not be ideal, I'll fix that.

tests/MockDb.h
```diff
@@ -244,6 +245,22 @@ struct MockDbT : public sqlpp::connection
   {
     return {name};
   }
+
+  void start_transaction(sqlpp::isolation_level level)
+  {
+    _current_isolation_level = level;
```
rbock: Storing the isolation level in the connection (even the mock) gives the wrong impression, I think.

rbock:
Thanks!

tests/MockDb.h
volka: Yes, that does not look nice, but I think this test is useful to verify that the correct internal start_transaction method gets called with the expected level.

I added an extra "InternalMockData" and moved the isolation_level there to make it explicit that this is only for testing.

Does this look better?

rbock:
Actually, looking at the slite3 connector, I might have been wrong about this. If I read the documentation correctly, the isolation level is set per connection, not per transaction.

If that is correct, then the isolation level should indeed be a member of the connection (to be able to query it) and there should be a `get_isolation_level` and `set_isolation_level` method instead of just passing the parameter to `start_transaction`.

Or we need to reset the isolation level upon ending the transaction. But that might turn out pretty ugly...

PR #172, opened 2017-06-02
ddl2cpp: Support implicitly treating id column as auto-incrementing (e.g. for SQLite)

AndiDog:
See https://sqlite.org/autoinc.html

This parameter for ddl2cpp allows us to use the performance of SQLite's implicit ROWID without having to declare the `id` column as `AUTOINCREMENT` (which is slower because SQLite ensures to not reuse deleted values). As effect, a primary key column called `id` is treated as automatic, i.e. does not need to be filled on insert.


rbock:
Cool, thanks, Andi!

PR #179, opened 2017-08-09
Fixed mising namespace comment on tablenames (to silence clang-tidy).

sergeken:
This fixes clang-tidy complaining about missing namespace closing comments

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -318,7 +318,7 @@ for create in tableCreations:
             traitslist.append(NAMESPACE + '::tag::require_insert')
         print('      using _traits = ' + NAMESPACE + '::make_traits<' + ', '.join(traitslist) + '>;', file=header)
         print('    };', file=header)
-    print('  }', file=header)
+    print('  } // namespace ' + tableNamespace, file=header)
     print('', file=header)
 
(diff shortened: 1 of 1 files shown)


rbock:
That sounds like a very strange rule in clang-tidy.
Also, there are mo namespace closing comments anywhere in sqlpp11. Shouldn't this rule lead to a whole explosion of complaints?

Just being curious.

sergeken:
It's one of the readability rules (part of a larger pack) that can be activated.  I do have it as my personal style also, as I actually don't like the C-style accumulation of of braces.  I've had troubles with imbalances when doing large C code refactorings in the 80's on dumb terminals.  Unfortunately all newer languages tend to follow the C style and disregard the cleaner (but more verbose) pascal style.

To come back on your remark about not using closing comments, you do actually have one in the ddl2cpp on line 341.  But you're right you don't do it in the include part of sqlpp11.  The reason why clang-tidy didn't complain about it, is because it considered these as part of the "system " headers (/usr/local/include) and errors/warnings are muted by default for these.

I leave it up to you to decide to accept this change or not for ddl2cpp.

rbock:
Thanks for the explanation and pointing out the already existing comment :-)

PR #180, opened 2017-08-10
add missing <string> include in schema.h

theodelrieu:
Hello, I had a compiler error while building the conan package on macOS.

There is a missing string include in `sqlpp11/schema.h` that I've added.

The bug can be fixed by reordering includes in the client code too.


rbock:
Thanks

PR #188, opened 2017-10-11
surround table name and column names with quotes (` `) in case they h…

shuangye:
…ave the same name with SQL keywords

surround table name and column names with quotes (` `) in case they have the same name with SQL keywords, e.g. `order`.


rbock:
With all CI checks broken, I am closing this pull request.

Please also confer #199.

PR #195, opened 2017-10-21
Allow result_t to have a size() function

Erroneous1:
Allow result_t to have a size() function if DbResult has a size() function.

size() must be const, but can be arbitrary type. If needed we could change this in the future.
Calling size() after popping from result is undefined behavior. Call size() first.

If your connector doesn't support size(), start a transaction, get select count(), then run select, and finish transaction. I assume the count won't change if you first start a transaction.

A proper size() function is necessary for doing things like vector<>.resize() and vector<>.reserve().

We may wish to move the helper structs for determining whether or not a result has a size() function to type traits.

Tested on Visual Studio 2017 using LLVM toolset.

### DIFF
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -38,6 +38,31 @@ namespace sqlpp
     using type = std::input_iterator_tag;
   };
 
+  namespace detail
+  {
+	template<class DbResult, class = void>
+	struct result_has_size : std::false_type {};
+
+	template<class DbResult>
+	struct result_has_size<DbResult, void_t<decltype(std::declval<DbResult>().size())>>
+	  : std::true_type {};
+
+	template<class DbResult>
+	constexpr bool result_has_size_v = result_has_size<DbResult>::value;
+
+	template<class DbResult, class = void>
+	struct result_size_type { using type = void; };
+
+	template<class DbResult>
+	struct result_size_type<DbResult, void_t<decltype(std::declval<DbResult>().size())>>
+	{
+	  using type = decltype(std::declval<DbResult>().size());
+	};
+
+	template<class DbResult>
+	using result_size_type_t = typename result_size_type<DbResult>::type;
+  }
+
   template <typename DbResult, typename ResultRow>
   class result_t
   {
@@ -139,6 +164,13 @@ namespace sqlpp
     {
       _result.next(_result_row);
     }
+
+	template<class Size = detail::result_size_type_t<DbResult>>
+	Size size() const
+	{
+	  static_assert(detail::result_has_size_v<DbResult>, "Underlying connector does not support size()");
+	  return _result.size();
+	}
   };
 }  // namespace sqlpp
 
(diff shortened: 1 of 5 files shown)


include/sqlpp11/result.h
```diff
@@ -38,6 +38,31 @@ namespace sqlpp
...
+  {
+	template<class DbResult, class = void>
+	struct result_has_size : std::false_type {};
+
+	template<class DbResult>
+	struct result_has_size<DbResult, void_t<decltype(std::declval<DbResult>().size())>>
+	  : std::true_type {};
+
+	template<class DbResult>
+	constexpr bool result_has_size_v = result_has_size<DbResult>::value;
```
rbock: Template variables are available in C++14, not in C++11.

rbock:
Nice!

Please remove the template variable to allow C++11 compilation.

Erroneous1:
I wonder if the PostgreSQL connector could implement a size() function by quickly iterating over the results (without binding) and counting, then seek to 0. IDK if SQLite3 has the ability to rewind/reset the results back to the first row.

Maybe it'd be better to have a different function like count_results() or something to indicate that this is not necessarily a quick and easy call.

Erroneous1:
Should I add another MockDB that doesn't have a size() function to make sure the static_assert never fires if you don't try calling result_t.size()? I know it worked fine in my environment, and if I understand templated defaults correctly the size() function shouldn't try to compile until you write code that calls size() and it then tries to figure out what the Size type should be (unless specified) and then it runs the static_assert. I wonder if void_t will work to prove a negative when a function definition is well formed but fails static asserts. I doubt it.

rbock:
MockDB: That would be great!
PostgresSQL/Sqlite: The alternative function sounds more reasonable, although maybe people should just `select count(whatever)` instead?

Erroneous1:
The only issue with select count() is race conditions if they don't get a read lock on the tables between selecting the count and then selecting the results. Still if you just need an estimate or don't expect the count to change then should be fine. I'm betting when you have SQLite3 as your connector you generally don't expect the race condition to happen.

It looks like you already have a `_handle->_handle->result.records_size()` and `_handle->totalCount` on PostgreSQL connector. If you could move setting up the totalCount to a constructor then a size() function could just return that or just make totalCount mutable so it can be modified within `size() const`.

rbock:
Right, race conditions might be an issue. Regarding PostgreSQL, that is not my code, but a good find for sure.
But this indicates that we really should start with the size function and adjust the connectors that can provide this information. Afterwards, we can think about other ways of counting (if required).

In other words: With a second MockDB to test the static assert, we could merge this and make the size() function available for connector libraries.

Erroneous1:
I removed size() from MockDbT and created a special MockSizeDb to test that size() works. Everything checks out on Travis and AppVeyor (except unrelated known issues with MS VS 2017). I also updated sqlpp-connector-odbc to implement size().

rbock:
Nice work! Thanks!

Issue #186, opened 2017-10-11
Cannot handle column names that match SQL keywords.


shuangye:
I have a column named "order", which is also a SQL keyword. When I execute a select, it throws an exception:
terminate called after throwing an instance of 'sqlpp::exception'
  what():  Sqlite3 error: Could not prepare statement: near "order": syntax error (statement was >>SELECT Areas.uuid,Areas.order FROM Areas<<
Putting the table name and column names in a quote (\') works fine: SELECT \`Areas\`.\`uuid\`,\`Areas\`.\`order\` FROM \`Areas\`.
But I have not find where the SQL statement is constructed in sqlpp11.

rbock:
Hi,

Yes, as of now, column names are not quoted. You could change that in column.h (the serialization function is at the bottom). Want to experiment with that and maybe create a pull request?

Cheers,

Roland

shuangye:
Thank you. I've added the quotes and tested. Pull request has been submitted.

Erroneous1:
This has been discussed before with issue #152 . The issue with quoting all tables is that some databases do not support backticks to quote names. The advice is to put the backticks in your column definition.

shuangye:
Right, double quotes works. 
```
CREATE TABLE Test("order" INT);
SELECT "order" FROM "Test";
```
Anyway, column.h needs to be modified to include such a mechanism to escape SQL keywords.

rbock:
@Erroneous1 Thanks for the reminder! I need to add this to the documentation, including the link to stack SO. That is one awesome answer!

@shuangye Erroneous1 is right, the quoting should be done at the source, which is in the table definition (sorry for leading you towards column.h yesterday, obviously not thinking clearly). The question is how to do it? I would prefer not to have a long list of potential keywords (which might differ from database to database). Maybe quotes should be read from the DDL?

Erroneous1:
According to the SO answer, ``` ` ```, `"`, and `[ ]` are the 3 ways to do it. It sounds like not all RDBMs support `"`, and it could be confused with text. A column could have a static constexpr enum or tag lets the serializer put quotes as either none (default), backtick, double_quote, or bracket.

Maybe something along these lines:
```c++
namespace quote_tag
{
struct none {};
struct backtick {};
struct double_quote {};
struct bracket {};
}

template<class Column, class = void_t<> >
struct requires_backtick : std::false_type {};

template<class Column>
[...]

PR #197, opened 2017-11-09
ddl2cpp: Escape column names if they are reserved keywords

dirkvdb:
I just ran into issues related to reserved column names (#152)

I made a change to the ddl2cpp script to quote table rows that are reserved sql keywords.
I only added my two conflicting keywords to the list though, this should be extended to all the sql keywords.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -48,6 +48,16 @@ def get_include_guard_name(namespace, inputfile):
   val = re.sub("[^A-Za-z]+", "_", namespace + '_' + os.path.basename(inputfile))
   return val.upper()
 
+def escape_if_reserved(name):
+    reserved_names = [
+        'GROUP',
+        'ORDER'
+    ]
+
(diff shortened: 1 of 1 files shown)


rbock:
Please also look here:

https://github.com/rbock/sqlpp11/issues/186 and the link to SO:
https://stackoverflow.com/questions/10573922/what-does-the-sql-standard-say-about-usage-of-backtick/10574031#10574031

There are multiple ways of quoting. And of course, there are multiple sets of keywords.

The current commit will break MySQL, for instance. It would be possible for the script to indicate that the name has to be quoted and then the connector would have to provide a function for quoting names.

I suggest closing this pull request. I will consolidate the existing tickets for quoting.

I'd be happy to see a new pull request to fix the problem (I will tackle this eventually, but it will take a bit of time).

dirkvdb:
I agree to close, especially if it breaks mysql. I will wait for a proper fix. I can manage with my version of ddl2cpp for now.

PR #215, opened 2018-02-08
Add the possibility to keep the table and column names as in the DDL

### LINKED ISSUE #213: Do not camelCase table and column names in ddl2cpp
I was wondering the reasons for transforming the table and column names in the ddl2cpp script. The `toClassName` and `toColumnName` specifically will [Cc]amelCased the original names so that a table `my_table` and `my_column` will have a class and member representation of `MyTable` and `myColumn`.
I find it weird and would prefer using the actual DB names in my code.

* Are there any reason for this behavior?
* Would you be open to change it or add an option to the script to get verbatim DB names in C++?

sjoubert:
Close #213


rbock:
Merci beaucoup!

PR #218, opened 2018-02-13
Allow digits 0-9 in include guards

alnr:
Without this change, headers generated for tables `foo0` and `foo1` in the same namespace have clashing include guards.


rbock:
Thanks!

PR #219, opened 2018-02-15
add set(tuple<...>) implementation for insert and update statements

### LINKED ISSUE #216: passing tuple into 'set' statement
Hi!

I'm not an expert in using your lib, so I started to wonder if there is a way to pass std::tuple of arguments into set statement, instead of a bunch of arguments.
to instead of doing 
```
struct Password {std::string value, function, salt; } pass;
struct Address { std::string city, std::string street; } address
db(insert_into(t_auth_info)
  .set( //
  t_auth_info.password_hash   = pass.value,
  t_auth_info.password_method = pass.function,
  t_auth_info.password_salt   = pass.salt,
  t_auth_info.address_city    = address.city,
  t_auth_info.address_street  = address.street) },
) 
```
just do

```
auto password(Password pass){
  return make_tuple(//
  t_auth_info.password_hash   = pass.value,
  t_auth_info.password_method = pass.function,
  t_auth_info.password_salt   = pass.salt);
}

auto address(Address address){
  return make_tuple(//
  t_auth_info.address_city    = address.city,
  t_auth_info.address_street  = address.street );
}

db(insert_into(t_auth_info).set(password(pass), address(address)))
```
or
``` 
db(insert_info(t_auth_info)).set( std::tuple_cat(password(pass), address(address)) )
```

Or is there some other way of achieving similar functionality?

cszawisza:
resolves #216

### DIFF
--- a/include/sqlpp11/update_list.h
+++ b/include/sqlpp11/update_list.h
@@ -258,7 +258,7 @@ namespace sqlpp
           -> _new_statement_t<check_update_static_set_t<Assignments...>, update_list_t<void, Assignments...>>
       {
         using Check = check_update_static_set_t<Assignments...>;
-        return _set_impl<void>(Check{}, assignments...);
+        return _set_impl<void>(Check{}, std::make_tuple(assignments...));
       }
 
       template <typename... Assignments>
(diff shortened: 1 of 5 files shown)


include/sqlpp11/update_list.h
```diff
@@ -279,7 +287,15 @@ namespace sqlpp
           -> _new_statement_t<consistent_t, update_list_t<Database, Assignments...>>
       {
         return {static_cast<const derived_statement_t<Policies>&>(*this),
-                update_list_data_t<Database, Assignments...>{assignments...}};
+                update_list_data_t<Database, Assignments...>{std::forward_as_tuple(assignments...)}};
+      }
+
+      template <typename Database, typename... Assignments>
+      auto _set_impl(consistent_t /*unused*/, std::tuple<Assignments...> assignments) const
```
rbock: gcc on Travis seems to be confused by this version of `_set_impl` and the "old" one when facing a single assignment.

Smells like a compiler bug to me.

I believe that we could work around this by keeping only this version (i.e. call `make_tuple` in `set` in case of individual assignments.

rbock:
Very nice, thanks!

PR #220, opened 2018-02-19
Install and export ddl2cpp script

sjoubert:
This patch install and export the ddl2cpp utility script as a CMake target.

Two remarks: 
* I've renamed, on install, the script to sqlpp11-ddl2cpp as I found the name not specific enough to avoid potential conflict. I've not renamed it in the sources to avoid breaking users that may rely on that in a cloned repo. Feel free to suggest another approach or another name.
* I didn't find a nicer way to export/import the associated target. I handled it by hand. Especially, I copied the checks generated by CMake for normal exported targets (Is the target already exported?, Does the script file exists?). I didn't find a way to make CMake export the utility scritp as a target without "building" it. If you have a solution I'll gladly fix that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,11 @@ install(TARGETS sqlpp11
         EXPORT Sqlpp11Targets
 )
 
+install(PROGRAMS "${PROJECT_SOURCE_DIR}/scripts/ddl2cpp"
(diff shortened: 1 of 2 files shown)


rbock:
Thanks! Looks good to me (not being a CMake expert, though :-) )

PR #221, opened 2018-02-23
Add more PostgreSQL type mapping in ddl2cpp

sjoubert:


matthijs:
Now every type for postgresql is defined here.
Is it possible to extend this script in the connector library and put there the postgresql specific stuff?

sjoubert:
I've added the missing types to match the PostgreSQL connector script with the exception of `USER-DEFINED` type as I found it weird. @cszawisza should I add still add it?

@matthijs I don't know what will think @rbock
For the time being I'm good with this addition in the main repo but it may make sense to have a customization point for each connector though it will probably require some design and update to the script.

cszawisza:
@sjoubert to be honest I don't know why i add `USER-DEFINED` in the first place... so removing it is a good idea :)

matthijs:
@cszawisza, @sjoubert 

I do, in PostgreSQL you can define your own types and when fetching the schema from the information_schema table you get as type 'USER-DEFINED'. That's why it is added to that script.

When a sql dump is made from the PostgreSQL schema you won't get that 'USER-DEFINED' type, but the actual (your) type. The current script parses a SQL dump so there is no need to have 'USER-DEFINED' there anymore. Removing it is ok here.

rbock:
Having special options for individual connectors sounds like a good idea to me.

I am a bit occupied with another project right now, but it seems to me that you are collectively steering in a good direction :-)

sjoubert:
I don't really have a lot of time either, would you still consider merging this PR and I can open an issue to track further improvements.

rbock:
Right, for now this is better than not merging.

PR #223, opened 2018-03-07
Fix blob result_type

blastrock:


blastrock:
I added support for std::array, I hope I didn't miss anything ^^

So, do you think we could merge this feature to master? Or is there still work to do?

rbock:
Nice work! Thanks :-)

Will merge into master, soon.

rbock:
Merged to develop now. Can you check if it works OK? (Just to be sure, I had a few merge conflicts).

blastrock:
develop works for me, with my branch on sqlpp11-connector-sqlite3

PR #225, opened 2018-03-09
prefix target sqlpp11 with namespace sqlpp11::

theodelrieu:
The main idea here is to have a way to group sqlpp related targets.
e.g. sqlpp11::sqlpp11-connector-sqlite3


rbock:
Being super ignorant about cmake, I admit, I have no idea if and what kind of implications this would have.
Closing this for now.

PR #227, opened 2018-04-09
sqlite3: added real data type

snikulov:
Added support for SQLite3's [REAL](https://www.sqlite.org/datatype3.html) data type


rbock:
Thanks!

PR #231, opened 2018-04-18
ddl2cpp: support CHECK constraint

th0br0:
The current ddl2cpp does not support parsing of CHECK constraints, failing on valid create table such as:

```
CREATE TABLE ex1 (
  id INTEGER,
  CHECK (id > 0 and id not null)
);
```

This PR adds support. Note that it does not support recursion of math expressions due to the complexity that this would add to the parser.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -178,7 +178,16 @@ negativeSign = Literal('-')
 ddlNum = Combine(Optional(negativeSign) + Word(nums + "."))
 ddlTerm = Word(alphanums + "_$")
 ddlName = Or([ddlTerm, ddlString])
+ddlMathOp = Word("+><=-")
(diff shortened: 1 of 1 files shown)


rbock:
Thanks!
Seeing this, I am quite surprised that nobody mentioned a problem with CHECK before...

egorpugin:
Hi!

I have one idea about parsing schemas.
At least this might work with sqlite.

Algorithm:
1. create dummy db with input schema
2. extract tables and columns with pragmas (other DBMS will use select queries to service tables)
3. simple parsing if needed

Pros:
1. You won't find (and write) a better parser rather than DBMS' built-in.

Cons:
1. Maybe it will be hard to extract some info - but this is unlikely.

rbock:
Hi,

That is a very nice idea, especially for sqlite.

With others it might complicate things because you then need the right to create temporary databases for building a project.

Cheers,

Roland

PR #233, opened 2018-04-20
Add simplified sqlite2cpp generator script that uses sqlite3 schema parsing.

egorpugin:
Later printing could be generalized between both scripts.
Reference: https://github.com/rbock/sqlpp11/pull/231#issuecomment-382505274


rbock:
Sorry for taking so long...

Looks pretty neat!

PR #249, opened 2018-09-25
Add option to ddl2cpp to enable splitting tables

Erroneous1:
`-split-tables` places each table into its own .h using path to target as a directory and sql table name as the filename.


rbock:
Nice option :-)

Thanks!

PR #251, opened 2018-10-26
Fix blob field display

fjkfwz:


rbock:
Thanks.

Please provide a small description next time you provide a pull request. It makes reviewing simpler and the pull request is likely to get in quicker :-)

PR #253, opened 2018-12-06
Add explicit greater condition.

egorpugin:
New VS2019 Preview (MSVC 19.20) shows following warning without this fix:
```
warning C4305: 'if': truncation from 'size_t' to 'bool'
```


rbock:
Thanks!

Issue #259, opened 2019-01-23
Support for std::string_view text operands


dermojo:
Hi,
Currently, sqlpp11 requires text operands to be implicitly convertible to std::string, which doesn't work for C++ 17's std::string_view.
I have a local fix to add overloads to text_operand and to adapt the check in wrap_operand.
Does it make sense to submit a patch or would this be superfluous with sqlpp17?
Thanks

rbock:
Hi,

With sqlpp17, this would work out of the box.

A patch for sqlpp11 might be nice to have nonetheless, although I am slightly afraid of the additional test overhead (we would need to have test with both C++11 and C++17).

Best,

Roland

dermojo:
Hi,

Yes, that would double the tests on travis/appveyor, but I think it's worth running the tests with multiple compilers/compiler versions and C++ versions.

I'll create a pull request for string_view.

Thanks,
Daniel

rbock:
Hi Daniel,

OK, looking forward to it.

Cheers,

Roland

PR #260, opened 2019-01-24
support for std::string_view when using C++17

dermojo:
See issue #259

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,12 +38,11 @@ before_script:
          sudo apt-get install python-pyparsing;
      fi
    - if [[ "$CXX" = "g++" && "$CONFIG" = "Debug" && "$TRAVIS_OS_NAME" = "linux" ]]; then export CXXFLAGS="--coverage"; fi
-   - cmake .. -DCMAKE_BUILD_TYPE=$CONFIG -DCMAKE_PREFIX_PATH=$PWD/../date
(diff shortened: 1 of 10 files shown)


dermojo:
VS2017 doesn't build, but the errors look unrelated to my changes...

rbock:
Nice work.

Can you add a test?
Ideally, also adjust `.travis.yml`.

Thanks,

Roland

dermojo:
Thanks.
I tested a few things and the easiest/best way to compile for C++17 is to allow to - optionally - specify a C++ standard in the CMakelists.txt. This can then be selected from the command line, e.g. when building with travis.
I think I will add this tomorrow.

dermojo:
I adapted the CMakeLists.txt for all tests to allow building them with an arbitrary C++ standard.
This cannot be done for the sqlpp11 library itself (it's an INTERFACE target), but for all tests using it.

Also, travis now builds the release version twice: once like before and once with C++ 17.
Currently, the C++17 only do something different on OS X, because the clang and gcc versions on the Linux image are too old, so there's still some potential for optimization...

rbock:
Nicely done! Thanks for the effort!

(sorry for the late answer, I was offline for a few days)

dermojo:
Great, thanks a lot for a very interesting library!

PR #261, opened 2019-02-27
cmake improvements and bugfix with unsigned fields

fraillt:
Hi, 
First of all, I would like to thank for this awesome library, and I have to suggest a few improvements and bugfixes.
1) add "Sqlpp11" namespace to CMake target
RATIONALE: This is conventional in modern CMake, to have namespace the same name as a package name. e.g.:
```cmake
find_package(Sqlpp11)
...
target_link_libraries(mylib PRIVATE
  Sqlpp11::Sqlpp11) 
```
It should be noted, that main "sqlpp11" target should be changed to "Sqlpp11" too. (but I didn't do this in this PR)
2) do not change HinnantDate_ROOT_DIR after a library is found
RATIONALE: Do not change constants provided for configuration, during configuration. It fails in more complicated projects, that has multiple subprojects that each searches for the same libraries.
3) Do not show error, if a target is already defined via find_package.
RATIONALE: same logic applies as in 2), and if a target is already defined, the best option is to do nothing.

Regarding the bug for unsigned field types, I had an issue in this SQL.
```cpp
  _db->run(sqlpp::update(playerAccount)
    .set(playerAccount.curProxyId = sessionInfo.proxyId,
      playerAccount.loginCount = playerAccount.loginCount + 1,
      ...)
   ...
```
"+ 1" simple didn't compile :) i find that when loginCount is signed int, then it works, so i applied same logic to unsigned expression operations as in signed ones, but I'm not very confident with this fix. :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,6 +98,7 @@ set(ConfigPackageLocation lib/cmake/Sqlpp11)
 install(EXPORT Sqlpp11Targets
   FILE
     Sqlpp11Targets.cmake
+  NAMESPACE Sqlpp11::
   DESTINATION
     ${ConfigPackageLocation}
 )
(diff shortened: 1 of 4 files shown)


fraillt:
Regarding H.Hinnant's date library, it has cmake support for quiet some time now, I would strongly suggest to remove find module for it

rbock:
Hi,

Thanks for the pull request. the change for unsigned integral looks good. 

I am still a bit illiterate when it comes to cmake, But I appreciate anything that makes the cmake files more canonical.
Given the cmake support in Howard's date library, I also guess you are right that we should drop the local cmake file.

In any case, please also adjust the travis and appveyor configuration (and maybe the README.md) to work correctly with the new cmake files.

Thanks,

Roland

rbock:
I adopted the changes to unsigned integral values.

As all the CI builds failed with the proposed changes in the cmake files, I am closing this pull request.

PR #280, opened 2019-07-09
Add -identity-naming argument from ddl2cpp to sqlite2cpp.py

bboston7:
This pull request ports the `-identity-naming` argument from `scripts/ddl2cpp` to `scripts/sqlite2cpp.py`.  The change preserves the current behavior of `sqlite2cpp.py` by default, but when the script is called with `-identity-naming` as an argument it will emit a .h file that preserves the input DDL's case of the table and column names in the generated output.


rbock:
Nice work, thanks!

PR #298, opened 2019-10-14
Use Python3 for test_scripts

uyha:
Update test_scripts to use Python3 instead of FindPythonInterp (deprecated since CMake 3.12). Ignore Python2 since its support's gonna dropped soon also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ notifications:
 
 install:
   - g++ --version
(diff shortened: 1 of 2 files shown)


rbock:
The travis image currently in use has cmake 3.9

I am in favor of upgrading to a more recent version of travis, your help would be welcome.

rbock:
NIce work! Thanks!

PR #306, opened 2019-12-03
Make sure find_package(Sqlpp11) can be done multiple times

sjoubert:
It is expected that CMake packages can be found multiple times without
errors. The target generated and defined by CMake, e.g. in a
*Targets.cmake file have a similar guard.


rbock:
Nice work. Thanks!

Issue #315, opened 2020-02-13
Support for OVER() clause


MacDue:
Would it be possible to get basic support for the OVER() clause at some point?
I'd help (me) out with doing some simple pagination. 

```sql
SELECT hp, COUNT(1) OVER() as total_quests 
from active_quests 
LIMIT 3 OFFSET 1;
```
I'd expect the sqlpp syntax to be something like:
```cpp
 select(active_quest_tbl.hp, count(1).over().as(total_quests))
.from(active_quest_tbl)
.limit(3u)
.offset(1u);
```
I'm interested in implementing it myself, I see that I'd need to implement an expression type `over_t` & an operator `.over()` but I'm kinda unsure on the details of implementation (and stuff like limiting it aggregate functions).

rbock:
Hi,

Wow, I wasn't even aware of `OVER` until this morning. Thanks for pointing me to it :-)

The short answer is: Yes, it would be possible.
See for instance here: https://github.com/rbock/sqlpp11/blob/6d2b64eff9ecbed300cf8c7d2a50c5dd87ca7d71/include/sqlpp11/aggregate_functions/count.h#L59

As of now, I see two options (there are probably more):

- You can follow this style and let the aggregate functions inherit from another CRTP base class that provides the `over` member function.
- You can add a one-line  `over` member function to each of the aggregate functions, which dispatches to a free `over` function.

In both cases, the `over` function would return a new `over_t<Expression>` object. 

These days I am leaning towards the second option and would start with that free function.

Does that help to get you started? Let me know if you get stuck.

Cheers,

Roland

MacDue:
Thanks, I managed to get a basic implementation working :).
I went with the first method as I preferred the resulting syntax a little more -- a closer match to the sql.
#316 

There are more features to over that I've not implemented yet though (that could be tackled in the future too).

PR #316, opened 2020-02-14
Minimal over() implementation for aggregate functions

MacDue:
This PR adds minimal support for the `OVER()` clause (i.e support for `OVER()` with no arguments).
There are some more advanced features yet to be implemented described here: https://www.sqlservercentral.com/articles/understanding-the-over-clause

There implemented syntax is as follows:
```cpp
 select(active_quest_tbl.hp, count(active_quest_tbl.id).over().as(total_quests))
.from(active_quest_tbl)
.limit(3u)
.offset(1u);
```
Which is serialized to:
 ```sql
SELECT hp, COUNT(id) OVER() as total_quests 
from active_quests 
LIMIT 3 OFFSET 1;
```

Note:
```cpp
 select(active_quest_tbl.hp, count(active_quest_tbl.id).over())
```
Also works, in this case the name of the selection will be the aggregate function name (count in this example). 

This addresses #315


MacDue:
I'll look into the error (seems like whatever it is is not an issue in gcc 9)

rbock:
Hi @MacDue 

Very nice work! Would be great if you could add a test or two in `test_serializer`.

Best,

Roland

MacDue:
@rbock https://travis-ci.org/rbock/sqlpp11/jobs/650559734?utm_medium=notification&utm_source=github_status

Are auto aliases not applied at this point in serialization?

MacDue:
Added a little test for serializing .over() both with & without an alias.

test_serializer/Over.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+ * OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "Sample.h"
+#include "compare.h"
+#include <sqlpp11/sqlpp11.h>
+
+SQLPP_ALIAS_PROVIDER(dueutil)
```
rbock: Just curious. Is this a typo? Or am I missing some reference?

rbock:
With the possible exception of the one comment, this looks very good to me.

Let me know if you think it is ready to be merged, too.

test_serializer/Over.cpp
MacDue: I just randomly put the name of the project I added this for the test. Don't mind changing it to something else if you like.

MacDue:
I think it's pretty much ready. Maybe some thought should be spared for the other features of `OVER()` at a later date, but it'd require quite a bit of extra work (and would probably just be adding stuff for completeness).

test_serializer/Over.cpp
rbock: Thanks for the explanation :-)

rbock:
Done, thanks again :-)

PR #317, opened 2020-02-15
Fix connection pool compilation error

MacDue:
Tiny fix for a error I got when playing round with the connection pool built in to sqlpp11.
It seems like the constructor for a connection might have changed since it was implemented.

I know connection pools are not really a supported feature, but It's better fixed than not :)


rbock:
:-)

PR #335, opened 2020-05-19
Allow control over dynamic where equality operators

retsimx:
In the current implementation of sqlpp, the add function on dynamic where statements chain together expressions with AND only. This merge request allows more control over dynamic where so that OR or AND may be used.

### DIFF
--- a/include/sqlpp11/interpretable_list_with_string.h
+++ b/include/sqlpp11/interpretable_list_with_string.h
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2013-2015, Roland Bock
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ *   Redistributions of source code must retain the above copyright notice, this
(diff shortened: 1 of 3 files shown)


rbock:
Thanks for your pull request!

I am curious: Are you aware of `sqlpp::boolean_expression`? It also allows you to dynamically construct an expression. And it will allow you to get braces right.

retsimx:
No I am not! Thank you for pointing me at this. I have perhaps fallen victim to the docs! I'll have a look and see if constructing a `sqlpp::boolean_expression` works for my case later today. If so, I'll close this PR and submit a wiki PR instead.

rbock:
A wiki PR would be great :-)

retsimx:
It certainly appears that chaining together `sqlpp::boolean_expression` will be fine for my use case. This PR is indeed entirely redundant. When I have a chance I'll try to figure out how to make a wiki PR and document a few examples.

rbock:
Thanks for the feedback. I am looking forward to for wiki PR (not sure how that works either, feel free to send me some markdown).

PR #339, opened 2020-06-16
Add some MYSQL types to the ddl2cpp script

greenbrettmichael:
decimal and numeric are essentially the same.
The max size for longtext IS 4.3 GB. I do not know if it would fit in the text implementation at full size (because I haven't read it) but I imagine it would work for most users even if not.


rbock:
Thanks!

Issue #348, opened 2020-08-03
Pull in the Date Time lib via cmake


Leon0402:
Hi,

I would like to suggest to pull in the date lib via FetchContent. This would make it a lot easier to use! 

I had a look at the lib and it seems like FetchContent is supported and actually the preferred way https://github.com/HowardHinnant/date/blob/master/CMakeLists.txt. 

Would you be open for that? I might could help you with this.

rbock:
Hi Leon,

Thanks for the suggestion. I was not aware of FetchContent before, but quickly glancing at the documentation, this sounds good, indeed.

I guess the next step after that would be to something similar for the connector libraries?

Leon0402:
Yeah that would have been in deed the next thing I suggested :)
It has one drawback though, it requires at least cmake 3.11 in order to be available. I would even recommend 3.14 because that simplified the usage of FetchConent by introducing FetchContent_MakeAvailable().
Cmake just released 3.18, so I think bumping up the cmake version to 3.14 would be justified.

If that's okay I would send in a pull request.

rbock:
3.14 is about 1.5 years old, which is not a lot, but I would hope that most users can cope (I am on 3.16).

Let's go for it and push people to update their packages where required :-)

Regarding examples: Sure, the tests are compiled and executed. And you could use the library to serialize SQL expressions (dialects are supported in the connector libraries, only, though).

Leon0402:
Great, I will work on it!

I just saw https://github.com/rbock/sqlpp11/issues/174, is there any progress? That might be relevant for the next step you mentioned. With fetchContent usage would be really easy then, because you basically just have to link then against the connector you need (linking against the core lib will be done automatically) and you can switch easily.
With find_package() the situation is similar, you can make the connectors components, so usage would look like
```
find_package(sqlpp COMPONENTS sqlite3 REQUIRED)
```
and again you would get a target of the sqlite3 connector in this case and it would have a dependency on sqlpp. 

In my opinion, we should work on a clean cmake structure in order to make this happen. Otherwise you will end up with a 500 line top level cmake file.

rbock:
Hi,

Regarding #174 : I don't think I will make this happen any time soon, although I still believe that it would be the right thing to do in the long run.

As for the cmake structure: This is no thanks to me, as I am very much a cmake illiterate, but thanks to a number of great contributors, who helped straighten this out :-)

I am looking forward to your changes!

PR #349, opened 2020-08-04
Feature/fetch content

Leon0402:
Hi,

as discussed in #348 my implementation to use fetch content instead. I've also added instructions on how to use the lib and two examples (one for fetch content and one for find package).

I also made a little bit more fetch friendly for example by including CTest (which will automatically provide a BUILD_TESTING cache variable) and automatically deactive test, if it's not the top level project.

The main.cpp is empty in both though, do you have a very short code using any of the methods from the lib? Just to verify that everything is working. I browsed in the tests, but they seemed a little bit too complicated for this simple example. It doesn't have to be a useful example.


rbock:
Thanks for the pull request! I will need a bit of time to do a more thorough review.

For main.cpp
`test_serializer/Insert.cpp` might be a good inspiration? 

`compare(__LINE__, insert_into(bar).default_values(), "INSERT INTO tab_bar DEFAULT VALUES");` could be a good one-liner (plus includes).

When quickly scrolling through, I noticed that you set the C++ standard to C++14? Ideally, it should still work with C++11.

Leon0402:
No worries, take your time. I will be also happy to explain you some thing, if something is unclear. Btw. where is the travis build, shouln't it show up here? 

Thanks for the example, I will have a look. I also wondered if I should include the two examples in the pipeline? As kind of an deployment test? I'm not a travis expert, but I think this would be possible and helpful. 

> When quickly scrolling through, I noticed that you set the C++ standard to C++14? Ideally, it should still work with C++11.

This is only the C++ Version of the examples, it was just randomly chosen :) But I agree, it would be best to set it to the minimum version possible for testing purposes. I will change it

Leon0402:
> For main.cpp
> `test_serializer/Insert.cpp` might be a good inspiration?
> 
> `compare(__LINE__, insert_into(bar).default_values(), "INSERT INTO tab_bar DEFAULT VALUES");` could be a good one-liner (plus includes).

The problem is that this needs compare.h, which needs MockDb.h ... I don't wont to depend on these files, because people just just use that sample as a quick test to see if everything is working (so use it outside of this repository). 
Anything else, which just depends on the official headers?

rbock:
OK, so how about this as a one-liner?

```
select(sqlpp::value(false).as(sqlpp::alias::a)));
```

Does not do anything, but if this compiles, we should be in good shape.

We could serialize **without** MockDb, if you want some runtime test, too.

rbock:
Re Travis: Thanks for noticing. They are here: https://travis-ci.org/github/rbock/sqlpp11/pull_requests and they failed. Strange that they are not shown.

Leon0402:
I think this should be ready except the fact that the url in fetchConent is still pointing to my repo. (So I will leave it a draft). When you approve the pull request I will change that to your repo and the development branch.

I tested both the find_package as well as the fetchConent approach on my Linux machine.

dependencies/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) 2013-2016, Roland Bock
+# Copyright (c) 2016 Christian Dávid
```
rbock, line 2: I suppose you should add your name and 2020 here instead?

dependencies/hinnant_date/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) 2013-2016, Roland Bock
+# Copyright (c) 2016 Christian Dávid
```
rbock, line 2: Same here.

rbock:
Looks great!

In addition to the two comments: Please add copyright and license to the new example files, too.

Leon0402:
Okay all done :)

rbock:
Great! Assuming that the checks pass (strange that travis still does not show up), I'll merge during the weekend.

Leon0402:
> Great! Assuming that the checks pass (strange that travis still does not show up), I'll merge during the weekend.

I saw that issue on other repos as well, a quick search found this
https://travis-ci.community/t/travis-ci-build-no-longer-showing-as-check-on-github-pull-request-page/6955/10

Perhaps you have to renew something?

rbock:
Thanks for the hint. Apparently, I needed to sign up to beta to migrate some of the repositories.

rbock:
I updated the readme to point to travis-ci.com now. Could you rebase, please? We could see if the travis build shows up again :-)

Leon0402:
Seems to work again :)

rbock:
Great! I would squash and merge using these comments, OK?

* Add date as a dependency, make fetch friendly
* Add example
* Add alias
* Update readme
* Add find_package example
* Update travis

Leon0402:
Sure, go ahead! :+1:

rbock:
Done.
I'll let this stew for a couple of days before tagging a new release.

PR #353, opened 2020-11-04
Check dependencies.

GCarneiroA:
Hi, @rbock and @Leon0402  thanks for your answers, this pull in this repository serves to ensure that no other part of the project has already added the date library and causes duplication problems, necessary because the connector also adds date in its codes and without this update there would be compilation problems . Further explanations can be found at https://cmake.org/cmake/help/v3.18/module/FetchContent.html#populating-the-content

Edit: This code is valid for cmake <= 3.18, for cmake> = 3.19 the proposed new model can be used, but I believe that we must ensure that it works for users of old versions.

@rbock would you like me to add the comment to the commit message ? The failure occurs because this repository has not yet been merged and it depends on this PR to work correctly.

@Leon0402  the same has already been done by the connector, you can check it out at https://github.com/rbock/sqlpp11-connector-mysql/pull/64

Anyway, I left a private repository with the links pointing to the previously updated repositories, so that you can test the codes, at the most, the PULLs sent correct the links to the correct repositories. Please see the test repository at: https://github.com/GCarneiroA/mysql-connector.git

To test:
```
    git clone https://github.com/GCarneiroA/mysql-connector
    cd mysql-connector
    mkdir build
    cd build
    cmake ..
    make -j
```

In your code, just use includes and link with sqlpp-mysql::sqlpp-mysql    :)

Here I made available a test repository, which uses these PULLs, currently with the links pointing to my own repositories, but they work to test correctly: https://github.com/GCarneiroA/sqlpp11_mysql_test/blob/master/dependencies/CMakeLists.txt


rbock:
Hi,

Thanks for the pull request. Can you add some description, please?
Same for https://github.com/rbock/sqlpp11-connector-mysql/pull/64 (which currently looks like it is failing).

Since you are modifying the code introduced by @Leon0402 and I am anything but a cmake expert, it would be great if Leon could take a look, too :-)

Thanks,

Roland

dependencies/hinnant_date/CMakeLists.txt
```diff
@@ -22,4 +22,8 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-FetchContent_MakeAvailable(date)
\ No newline at end of file
+FetchContent_GetProperties(date)
+if (NOT date_POPULATED)
+    FetchContent_Populate(date)
+    add_subdirectory(${date_SOURCE_DIR} ${date_BINARY_DIR})
+endif()
```
Leon0402, line 29: This does the same as FetchContent_MakeAvailable

Leon0402 (changes requested):

Leon0402:
Hi, 

thanks for the pull request. It's nice that you add fetchContent support for the mysql connector and I will comment on this if we figured out this pull request. 

Could you please explain, what you trying to achieve with the pull request? It seems like you replaced the new way of fetching dependencies with the old more verbose way? Have a look here: https://cmake.org/cmake/help/latest/module/FetchContent.html

dependencies/hinnant_date/CMakeLists.txt
GCarneiroA: In versions of cmake <= 3.18 it is necessary to make a check, because https://github.com/rbock/sqlpp11-connector-mysql/pull/64 also adds a library also added by mysql-connector, which causes cmake.

This verification is explained at: https://cmake.org/cmake/help/v3.18/module/FetchContent.html#populating-the-content and I believe that we should still use the old code to ensure that it works correctly for users cmake <= 3.18

Leon0402: This is true for versions of cmake <= 3.14 not 3.18 (see https://cmake.org/cmake/help/v3.14/module/FetchContent.html#populating-the-content). 

As cmake 3.14 was older than 1.5 years the decision was made to bump the version up to 3.14 (the latest ubuntu lts comes with cmake 3.16 support). See #348 

I therefore suggest you follow sqlpp11 and also the sqlp11-sqlite-connector and bump the version of the mysql-connector also up to 3.14

GCarneiroA: At https://cmake.org/cmake/help/v3.19/module/FetchContent.html#examples there are times when we should use FetchContent_Populate and not FetchContent_MakeAvailable, this is the case, as the intention is to make the sqlpp-mysql-connector is the master, just using it to link the final project as shown in the usage example.

pc

I also leave a print explaining this use, even so in version 3.19, I believe that this way is better explained.

Leon0402: I know that there are special scenarios, where you cannot use fetchContent_MakeAvailable, in particular when you need to do more than just fetching. The example you have here shows it very well: There is the need to copy an additional file, therefore you need the more fine control. 

But I don't see why you need this here. If you look in the code, you see it's just the usual pattern, nothing special going on. 

Declaring date twice, once in sqlp11 and once in the connector, is fine (as it actually is also a direct dependency of the connector) and in no conflict. In fact, FetchContent was built to handle such a scenario just fine. The date lib doesn't conflict and is also only downloaded once (and not twice). This also a reason why I use this dependency pattern this way. If you call all FetchContent_Declare() before FetchContent_MakeAvailable (as you did) dependencies which are shared, will only be downloaded once. 

Just to be sure, I also tried out your other pull request in the connector and built it against master (rather than this pull request) and it just worked fine. 
So please be more specific on what exactly your proposed solution does better exactly and how exactly it is achieved (i.e. what is FetchContent_MakeAvailable missing what you currently do, I don't see a difference). 

I also took a look at your example project, although I didn't try it out. But here: https://github.com/GCarneiroA/sqlpp11_mysql_test/blob/master/dependencies/CMakeLists.txt you would only want to fetch the connector, which automatically fetches and links sqlp11.

GCarneiroA: Yes, the main idea is that only the connector is used and he is responsible for downloading the sqlpp11 dependencies, that's why I did this PR, to ensure that there will be no conflicts in the use of the connector, which actually happened without this PR .

In some previous tests, when only FetchContent_MakeAvailable was used in this repository, conflicts occurred mainly on date, so, for this reason I made this PR, this guarantees that there will be no future errors, besides it will not cause regressions in the current codes.

Forgive me if I couldn't explain it better before, my English is still not very good, so I'm trying to improve it too, but I hope that, if I agree, it can be merged, it will make the sqlpp11 much easier, just use the connector wanted :)

Leon0402: Could you then provide the code or tests, where conflicts occured, so I can check what went wrong there? 

As I said, you current mysql pull request just works fine when built against sqlpp11 master. So it would be interesting to see what actually doesn't work right now.

The sqlite3 connector already is built this way (so you fetch the sqlite3-connector and it will figure out the rest). I didn't ran into problems so far. 

Just to explain a little bit better: Have a look at FetchContent_MakeAvailable Implementation -> https://gitlab.kitware.com/cmake/cmake/-/blob/v3.14.0/Modules/FetchContent.cmake#L1041 

As you can see it's the exact same code as here with one extra check at the end. When you look at the code in recent master here -> https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FetchContent.cmake#L1148  you see that the code has evolved and does extra checking, which prevents errors. 

So in general, as the cmake documentation says as well, it's very recommend to use always this macro if you can. It will do a lot more extra checks and might also be improved with newer versions. 

So in theory (!) there should be no need to change anything here, in fact it would be counterproductive and against modern cmake practises. But perhaps cmake has a bug you run into or something else happened, so let's find out what went wrong :-)

GCarneiroA: Hi, I had to format my computer, I'll prepare again here where the error was and send it soon, no problem :)

GCarneiroA: I did some tests and it's really easy for you to test it too, create a new project and use this:
```
FetchContent_Declare (mysqlpp
     GIT_REPOSITORY https://github.com/rbock/sqlpp11-connector-mysql.git
     GIT_TAG origin / dependencies
)
FetchContent_GetProperties(mysqlpp)
if (NOT mysqlpp_POPULATED)
    FetchContent_Populate(mysqlpp)
    add_subdirectory(${mysqlpp_SOURCE_DIR} ${mysqlpp_BINARY_DIR})
endif()
```

See that the repository is the origin / dependencies that is in the PR for review, you will see that it does not fetch, conflicts occur.

But, if you use this:
```
FetchContent_Declare (mysqlpp
     GIT_REPOSITORY https://github.com/GCarneiroA/mysql-connector.git
     GIT_TAG origin / master
)
```

You will see that it will work correctly, as this repository calls the repositories with the necessary corrections.

Leon0402: Hi, I couldn't quite follow your instructions. The repository  https://github.com/rbock/sqlpp11-connector-mysql.git with branch origin/dependencies doesn't exist. I could find such a branch in your fork, but it didn't use sqlpp11 master and therefore just worked fine. 

I think it might be easier to figure this out, when we work on the same code, so we talk about the same thing. 
I forked your test repo, see https://github.com/Leon0402/sqlpp11_mysql_test. And changed it, so it links against my modification of your connector pulll request (which just uses sqlpp11 master instead of this pr), see https://github.com/Leon0402/sqlpp11-connector-mysql. 
So I just used fetchContent_MakeAvailable everywhere. You can try it yourself. 

I invited you as a collaborator to the test project. So you can play around with it and find a situation where it doesn't work. Then just push it and I can try it out and see what the issue is :)

GCarneiroA: Sorry, we still have an impace on that. The main objective is https://github.com/rbock/sqlpp11-connector-mysql/pull/65 and after some tests I could verify that the PR https://github.com/rbock/sqlpp11/pull/353 maybe not in fact, however, the main https://github.com/rbock/sqlpp11-connector-mysql/pull/65 still needs to use FetchContent_Populate instead of FetchContent_MakeAvailable as suggested, it is easy for you to change and test it, if you switching to the recommended pro will see that the cmake will fail.

My suggestion is, can we pay more attention to https://github.com/rbock/sqlpp11-connector-mysql/pull/65 ?

Still, considering! 65 I would keep! 353 the same, just to stay the same.

GCarneiroA: After a few more tests I was also able to discover that only the final projects using sqlpp11-connector-mysql will actually have an error, so using FetchContent_Populate this error will disappear.

This PR provides a very simple way to use this library in a final project, just fetch and link to the final executable.

Therefore, due to the new tests I would like to suggest the following procedure:
     1. Merge https://github.com/rbock/sqlpp11-connector-mysql/pull/65
     2. Test with https://github.com/GCarneiroA/sqlpp11_mysql_test (ERROR)
     3. Merge https://github.com/rbock/sqlpp11/pull/353
     4. Test again with https://github.com/GCarneiroA/sqlpp11_mysql_test (OK)
    5. If you are not in compliance, reverse the commits.


If there are no errors in the PRs, the tests above must perfectly demonstrate the operation and the error found.

GCarneiroA: Also note that for some reason the travis build test fails, I recommend that you test locally to verify that it is working correctly.

Leon0402: I agree, let's concentrate on  rbock/sqlpp11-connector-mysql#65 first. The problem is there the same, but I hope we can reproduce it better there by just changing it to use FetchContent_MakeAvailable and then your test project should fail and we  can have a look why exactly it is failing. 

The failing travis build has to do with the cmake version, I got this error as well, when I updated sqlpp11 and the sqlite3 connector. Have a look at my comments over there.

PR #363, opened 2021-04-21
performance optimize , move make new database connection out of critical section.

byronhe:
performance optimize , move make new database connection out of critical section.


rbock:
Thanks for spotting and resolving!

PR #364, opened 2021-05-08
Created docs directory

The-EDev:
A compromise between Github's wiki and in-code comments. Should allow people to open PRs to add documentation. And later on use mkdocs to generate a documentation site for Github Pages.

For now all I did was take the wiki and make the relative links work.


docs/Planned-Features.md
```diff
@@ -0,0 +1,29 @@
+# Planned Features
```
rbock, line 1: I guess we should drop this one:

- dynamic_join and date/time types are available.
- I don't have plans to support conditional columns in sqlpp11

The-EDev: Maybe we could put the descriptions in `Planned-Features.md` and `New-Features.md` in new or existing pages. For example `Union` could be added to `select.md` and `Custom Query` should probably have its own page. Not sure how to apply that to planned features though, since the description seems uncertain.. might be best to just drop that one completely and rewrite the feature descriptions later.

What do you think?

rbock: Yes, I suggest dropping `Planned-Features` completely.

And I agree, the content of `New-Fetaures` should be distributed accordingly and the page dropped as well.

But we can easily do that in another PR.

rbock:
Btw, I updated the travis config to drop OSX tests. The build failures shown here were just artifacts of using an outdated config.

docs/Home.md
```diff
@@ -0,0 +1,25 @@
...
+* [Tables](Tables.md)
+* [Insert](Insert.md)
+* [Select](Select.md) <- You might want to read this first as an appetizer
+* [Update](Update.md)
+* [Remove](Remove.md)
+* [Functions](Functions.md)
+* [Prepared Statements](Prepared-Statements.md)
+* [NULL](NULL.md)
+* [New Features](New-Features.md)
+* [Planned Features](Planned-Features.md)
```
rbock, line 25: If you drop this, too, we are good to go, I think :-)

The-EDev: right, I forgot about this 😅

PR #394, opened 2021-11-17
Add Connector CMake Logic

Leon0402:
Similar to what I did in sqlpp17 I added targets and install logic for all connectors. 

Client Code with FetchContent 
```
FetchContent(...) 
FetchContent_MakeAvailable(...) 

target_link_libraries(myTarget PRIVATE sqlpp11::postgresql)
```

Client Code with findPackage 
```
find_package(Sqlpp11 COMPONENTS PostgreSQL REQUIRED)

target_link_libraries(myTarget PRIVATE sqlpp11::postgresql)
```

I tested both variants (with postgresql) and it seems to work fine. It's not building yet, but that seems to be a missing function I will add it in another PR.


rbock:
Hi Leon, Thanks for the PR.

One of the huge benefits of sqlpp11 and its connectors being header only now, is that the library can be "installed" just by copying a couple of header files into place. For instance, there is no need to have any shred of Postgresql on your system to install those headers. It is totally independent. And since it is just a couple of files, I would always go ahead and install them all.

You have dependencies for building and running tests.

It might be nice to have CMake components (not sure if that is the right term), that depend on the underlying connectors to be present. But we should either provide all of these components or none of them. They should not depend on what is currently available on the system.

Please adjust the documentation for cmake illiterates like me :-)
 - How to make use of the CMake components for individual connectors (it looks like I could say `add_target_link_libraries(... sqlpp11::MySQL)` 

MySQL is a good example since MySQL::MySQL is not a well known cmake thing outside of this library, while Postgresql::Postgresql is.

Leon0402:
I don't fully understand. Despite being header only, you cannot use them without having the proper dependencies installed. That's the rationale for making them a requirement. 

I'm not sure if I get this right, but maybe your point is about installation? So you want to be able to install everything once and worry about the dependencies later (when you actually use the library)? 
Then you're right that's not supported currently the way I did it. Probably because I would expect things to already work that are on my system :-) Or if you really want it this way and know what you're doing you can always just comment out the "find_package(MySQL)" and the others. 
If you think this should be a supported use case, I will think again about how to support both ways. 

I will add some documentation later outlying how to use the library the different ways :-)

rbock:
Sorry for being unclear.

Yes, it was about installation. Let's say I want to use the library with sqlite3 now. But then, a week later, I decide to migrate to postgresql. It should just work. Without needing to install sqlpp11 again. Or I can replace sqlite3 with sqlcipher, just by changing a couple of build flags.

That's one of the benefits of a header-only library. Dependencies are compile-time dependencies, not install-time dependencies.

Leon0402:
I changed it accordingly so installation always happens now. Therefore I also changed the option name to make more clear that they are just for tests. 

I first thought this wouldn't be possible at all, because we link against the targets `SQLite::SQLite3` for instance and I believed targets we link to actually have to exist. But this seems to be only a requirement for static / shared libraries, but not for interface libraries, which makes sense.

I also updated the examples and added some more details to the readme. 

Let me know if something is missing.

Leon0402:
I was wrong, it's not working. 

When I use `FetchContent` now without setting the variables, it will give an error message for my target linking against sqlpp as the targets `Postgresql::Postgresql` and others are not available, because it won't do the `find_package(Postgresql)`. 

So basically we need the find_package calls when used with FetchContent. 

I'll find another solution, which works for both hopefully.

rbock:
Sorry I can't be of help. Just a couple of thoughts:

If I'd be writing the cmake rules, I think I would look at two different scenarios.

1. Building and running tests
2. Using the library in another project

In the second case, I would assume that the outer project is responsible for providing the Postgresql::Postgresql, for instance. The main challenge would be: What to with mysql or mariadb, which are not directly supported by cmake.

But again (I once learned that one cannot over communicate): My cmake knowledge is rudimentary at best, so feel free to ignore this :-)

Leon0402:
Yes I agree with you, although I would split it up the second point. The library should support the most common form of usages, which are perhaps: "Raw" usage, findPackage, fetchContent and git submodule. 

"Raw" is basically always supported. And the other three can be categorized into two categories: 
- Installed on the system (findPackage)
- Added via add_subdirectory (git submodule & fetchContent) 

The requirements for findPackage you already summarized: 
- Possible to install everything without having the dependencies installed / avilable at that moment

For add_subdirectory: 
- Ready to use targets with all dependencies
- Every connector disabled by default (so nothing unecssary is done)

I therefore disagree a little bit with: 

> In the second case, I would assume that the outer project is responsible for providing the Postgresql::Postgresql, for instance. The main challenge would be: What to with mysql or mariadb, which are not directly supported by cmake.

In the best case the library handles that, it's just more convenient. Especially when used with FetchContent, people also kinda expect that (I do at least :-)). 
Depending on the concrete situation it might be necessary to add some logic to disable it as we saw in previous PR by other people.

Leon0402:
So I modified the PR (again) and hopefully found an acceptable solution. 

When you want install the connector target (findPackage) or use the connector target (fetchPackage) or run connector tests, you need to set `BUILD_<CONNECTOR_NAME>_CONNECTOR` to `ON. 

By default it will then also check whether the dependencies are installed. I figured for fetchContent it is better to make this the default as the error message are very confusing otherwise (It basically doesn't relate the error with sqlpp11 at all, because of how these INTERFACE targets work). 

For the use case described by you I added the `DEPENDENCY_CHECK` option, which defaults to `ON`. Setting this to `OFF` will skip all the `find_package` calls. Therefore this allows to install all connectors without having the dependencies installed at that time. 

Let me know if that's a good solution for all the different use cases sqlpp11 want to support!

(I also did some other small adjustments like using the targets for the tests. You might wanna review them to make sure I didn't make any error. At least tests all build successfully locally.)

rbock:
Thanks for your effort. Much appreciated!

We'll see if this fulfils all use cases :-)

PR #395, opened 2021-11-17
Unsigned Integer Handler for postgresql connector

Leon0402:
Followup of #394 (should be reviewed first), which adds unsigned integer handling to postgresql as we already have for sqllite3. 

Now my project is also fully working with the integrated connector postgresql here. Will try soon my other project with sqlite3.


rbock:
This seems to be one commit behind? Looking forward to merging the unsigned support as well.

Leon0402:
I think after 20 force pushes I got this fixed :-)  

Now only the actual change is part of the PR

rbock:
Thanks for the PR!

PR #398, opened 2021-11-25
deps: fixed condition for fetching date::date

snikulov:
Here the issue with FetchContent for managing dependencies when using sqlpp11 as subproject.

Let's assume that we use libraries as `add_subdirectory ` in main project

```
...
set(BUILD_TZ_LIB ON CACHE BOOL "" FORCE)
set(USE_SYSTEM_TZ_DB ON CACHE BOOL "" FORCE)
set(COMPILE_WITH_C_LOCALE ON CACHE BOOL "" FORCE)
add_subdirectory(3rdparty/date EXCLUDE_FROM_ALL)

add_subdirectory(3rdparty/sqlpp11)
add_subdirectory(3rdparty/sqlpp11-connector-sqlite3)
...
```
By adding date using `add_subdirectory` we inject `date::date` target into superproject.
But this target is not honored and `sqlpp11` fetching dependency without any knowledge about it.
So we'll have one dependency and second which will be fetched from hard-coded tag.
That's not what we expecting to see.

Using introduced `USE_SYSTEM_DATE` build option will not help, because we have no installed development package.

To fix this issue we can check for already existed target `date::date` and not fetch dependency.
This also will work for `find_package(date REQUIRED)` case. If it'll be found, then it also provide `date::date` target and not fetch from remote.

@Leon0402 @kovdan01 Thoughts?

Same issue with any backends (I use sqlpp11-sqlite-connector) which unconditionally fetching `sqlpp11` as dependency without any clue that it already included in superproject.

### DIFF
--- a/dependencies/CMakeLists.txt
+++ b/dependencies/CMakeLists.txt
@@ -23,7 +23,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
-if(NOT USE_SYSTEM_DATE)
+if(NOT USE_SYSTEM_DATE AND NOT TARGET date::date)
     include(FetchContent)
 
     FetchContent_Declare(date


Leon0402:
So basically you're issue here is that you want to use sqlpp11 (and others) with git submodule or git subtree, right?

Out of interest: Any specific reason you use git submodule and not FetchContent yourself? 

Especially following statement of you made me wonder:

> Same issue with any backends (I use sqlpp11-sqlite-connector) which unconditionally fetching `sqlpp11` as dependency without any clue that it already included in superproject.

I see why you might have date as a git submodule (if you use it yourself in your project), but why would you additionally have `sqlpp11` as a submodule? It's not very useful on it's own. Why not just have the connector and let it fetch `sqlpp11`? 

Or do you try to avoid fetchContent completely for some reason?

> be fetched from hard-coded tag

Side-note: You can actually override this when using fetchContent yourself. Just include it via fetchContent and use whatever version you want, it will override it. 

But in general I don't see any issue with that change, seems useful to me.

Regarding to the connector: Seems fine to me as well, but as of a recent change, sqlpp11 now includes all its connectors. The transition is not completely finished yet (see other PR), but it will be soon.

snikulov:
> Out of interest: Any specific reason you use git submodule and not FetchContent yourself?

Why should I use it? Any particular reason to depends on internet connection and not from local git clone?

rbock:
> Because it two projects as @rbock create it. Otherwise it should be in singel repo.

Note: I merged the connectors for mysql, sqlite3, and postgresql into sqlpp11 (develop).

Beyond that, I'll leave the cmake discussion to you (my cmake foo is minimal at best).

snikulov:
I'll use single repo with pleasure as soon as it'll be released/tagged  :)

Leon0402:
> Why should I use it? Any particular reason to depends on internet connection and not from local git clone?

Well git clone does need an internet connection as well :-)  Now you need the internet connection once for the generator step afterwards, not too much different. Then you can use options like `FETCHCONTENT_FULLY_DISCONNECTED`. Additionally you could even use CPM for project or system wide caching of the downloads. 
And you can always set `FETCHCONTENT_SOURCE_DIR_<uppercaseName>` as well and use local git repositories. 

I see though how that might be not as convenient as submodule for machines where you usually don't have internet at all.

The reasons for FetchContent (for me) are mostly ease of use. No extra options during clone. Dependency Management fully in cmake. And you do not need to care (usually) about dependencies like date or even sqlpp11 if you just use the connector with fetchContent. 
Your approach here, as far as I understood, is to just make every dependencies rescursively a submodule. That can be quite a lot I guess. 

But I don't want to argue or convience you. If submodule works for you that's great. And I definitely think sqlpp11 should support that flow.

Leon0402 (approved):

dependencies/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 
-if(NOT USE_SYSTEM_DATE)
+if(NOT TARGET date::date)
```
Leon0402, line 26: After thinking about it we might should make that an or condition with what we had before? 

It works this way currently, because the find_package call is before the `add_subdirectory(dependencies)` call. But that doesn't seem very robust.

Leon0402 (changes requested):

snikulov:
> Well git clone does need an internet connection as well :-)

No need to if you work at local network with product repository without any remotes to anywhere else

> After thinking about it we might should make that an or condition with what we had before?
> It works this way currently, because the find_package call is before the add_subdirectory(dependencies) call. But that doesn't seem very robust.

The whole point is to check if target already exists. So yes - `find_package` or `add_subdirectory` should be called first.
Otherwise your dependency through `FetchContent` will be injected.
So it's perfectly robust.

Leon0402:
> The whole point is to check if target already exists. So yes - `find_package` or `add_subdirectory` should be called first.
> Otherwise your dependency through `FetchContent` will be injected.
> So it's perfectly robust.

There is no need though to add this requirement. It could easily happen in the sqlpp11 main CMakeLists that the find_package(Date) happens after `add_subdirectories(dependencies)` because of some refactorings.
If we do `not (USE_SYSTEM_DATE or TARGET date::date)` the order of calls is unimportant, which could prevent potential errors in the future.

snikulov:
Sorry, I'm not getting your point.
Could you please provide me with information how to fix issue I've described in my PR?
Should I change `if` condition to `if  (NOT USE_SYSTEM_DATE OR NOT TARGET date::date)` ?
From my standpoint it's equivalent what I already suggested.

Leon0402:
> Sorry, I'm not getting your point. Could you please provide me with information how to fix issue I've described in my PR? Should I change `if` condition to `if (NOT USE_SYSTEM_DATE OR NOT TARGET date::date)` ? From my standpoint it's equivalent what I already suggested.

Should be `if (NOT USE_SYSTEM_DATE AND NOT TARGET date::date)` I believe. 

It works the same way as before, but now you should be able to swap: 
https://github.com/rbock/sqlpp11/blob/develop/CMakeLists.txt#L81
and 
https://github.com/rbock/sqlpp11/blob/develop/CMakeLists.txt#L77

So it's more robust for potential changes in the future :)

snikulov:
Ok. Done.

Leon0402 (approved):

rbock:
Thanks for the discussion :-)

PR #399, opened 2021-12-08
Add missing includes

AndiDog:
Those headers use `throw sqlpp::exception` but do not include the relevant header.

I had stumbled over that in `include/sqlpp11/mysql/bind_result.h` since my project failed to build, but that is already fixed. This PR fixes some remaining places.


rbock:
Hi Andi,

Long time to no see :-)

Thanks for the PR!

Cheers,

Roland

Issue #400, opened 2021-12-08
AddressSanitizer errors on macOS in MySQL query tests


AndiDog:
Reproduced like this on develop branch (e7590dc):

```
brew services start mysql
mysql -uroot -e 'CREATE DATABASE IF NOT EXISTS sqlpp_mysql;'

# Not the full command, just the maybe relevant snippet (in short, turns on address sanitizer and some other stuff I like to use)
cmake [...] -DBUILD_MYSQL_CONNECTOR=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=1 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_FLAGS="-g -Wno-inline -Wall -Wconversion -Wformat=2 -Wshadow -pthread -fPIC -Qunused-arguments -fsanitize=address -fno-omit-frame-pointer -ferror-limit=5 -std=c++14" -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address" -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=address" -DCMAKE_BUILD_TYPE=Debug

make -j8
make test
```

That leads to:

```
Running tests...
Test project /Users/asommer/dev/extern/sqlpp11/.build/develop
      Start  1: sqlpp11.core.usage.BooleanExpression
 1/46 Test  #1: sqlpp11.core.usage.BooleanExpression ...........   Passed    0.51 sec
      Start  2: sqlpp11.core.usage.CustomQuery
 2/46 Test  #2: sqlpp11.core.usage.CustomQuery .................   Passed    0.09 sec
      Start  3: sqlpp11.core.usage.DateTime
 3/46 Test  #3: sqlpp11.core.usage.DateTime ....................   Passed    0.08 sec
      Start  4: sqlpp11.core.usage.Interpret
 4/46 Test  #4: sqlpp11.core.usage.Interpret ...................   Passed    0.09 sec
      Start  5: sqlpp11.core.usage.Insert
 5/46 Test  #5: sqlpp11.core.usage.Insert ......................   Passed    0.09 sec
      Start  6: sqlpp11.core.usage.Remove
 6/46 Test  #6: sqlpp11.core.usage.Remove ......................   Passed    0.08 sec
      Start  7: sqlpp11.core.usage.Update
 7/46 Test  #7: sqlpp11.core.usage.Update ......................   Passed    0.08 sec
      Start  8: sqlpp11.core.usage.Select
[...]

PR #401, opened 2021-12-08
Support MySQL connect timeout option

AndiDog:
This is a major requirement for production-ready software. If no connection timeout is configured, it can take an OS-specific amount of time before an exception is thrown, in case the database host cannot be reached. For me, a combination of Linux / AWS RDS took 2 minutes. The change works in that setup, using a value of 5 seconds. Could not test locally on macOS due to https://github.com/rbock/sqlpp11/issues/400.


rbock:
Thanks! Looks good to me.

PR #412, opened 2022-01-06
add example code for `count(1)` in Functions.md

linrongbin16:
Add an example in Functions.md to explain `count(1)`.

This SQL is tested (I'm developing a python module with pybind11 and sqlpp11, so I test it in my python module):
```
>>> db.count_all_ip_proxy()
MySQL debug: Executing: 'SELECT COUNT(1) AS count_ FROM tb_ip_proxy'
MySQL debug: Constructing result, using handle at 0x104fe7e50
MySQL debug: Accessing next row of handle at 0x104fe7e50
4920
>>>
```

### DIFF
--- a/docs/Functions.md
+++ b/docs/Functions.md
@@ -35,6 +35,11 @@ for (const auto& row : db(select(tab.name, avg(tab.value))
 }
 ```
 
+Use `count(1)` for simply query row count:
+```C++
+unsigned long n = db(select(count(1)).from(tab).unconditionally()).front().count;
+``` 
+
 # Text Functions
 ## concat
 Just use the + operator :-)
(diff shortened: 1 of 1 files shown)


docs/Functions.md
```diff
@@ -35,6 +35,11 @@ for (const auto& row : db(select(tab.name, avg(tab.value))
 }
 ```
 
+Use `count(1)` for simply query row count:
+```C++
+unsigned long n = db(select(count(1)).from(tab).unconditionally()).front().count;
```
rbock, line 40: `count` currently represents a signed value, i.e. today it should be 

```c++
const int64_t n = ...
```

It probably would make sense to change [count.h](https://github.com/rbock/sqlpp11/blob/main/include/sqlpp11/aggregate_functions/count.h) from `integral` to `unsigned_integral`.

It would then be correct to say 

```c++
const uint64_t n = ...
```

linrongbin16: Hi @rbock , I have update the return type to `const int64_t`

rbock:
Thanks for helping with the docs!

Issue #418, opened 2022-01-30
[BugReport] The tools of ddl2cpp won't process PostgreSQL 'schema' and default function


vrqq:
See the example SQL below:
```
CREATE TABLE "public"."dk" (
  "id" int8 NOT NULL DEFAULT nextval('dk_id_seq'::regclass),
  "last_update" timestamp(6) DEFAULT now()
)
;
```

In Postgres, the ```AUTO_INCREMENT``` usually represent with function ```nextval``` to add a _sequence item_ in same schema.

The problem currently have:
* ```"public"."dk"``` should matched on ```ddlName```
* ```ddlAutoValue``` may matched on function ```nextval()```
* ```now()``` cannot matched in ```ddlColumn```

The way to fixed the first one (schema name match) below:
```py
def class_name_naming_func(s):
  s = s.replace('.', '_')
  return re.sub("(^|\s|[_0-9])(\S)", repl_camel_case_func, s)


def member_name_naming_func(s):
  s = s.replace('.', '_')
  return re.sub("(\s|_|[0-9])(\S)", repl_camel_case_func, s)

ddlName = Or([ddlTerm, ddlString, Combine(ddlString + "." + ddlString)])
```
Tested on PostgreSQL 14.

It seems like there has the code about the last problem but that may not work.
I am not good at Python, so regret to have no idea to fix others.

rbock:
Took a few days, but it should work now.

Note: I partially rewrote the code generator to fix this. My test cases work, but this is super fresh :-)

PR #419, opened 2022-01-30
Fix part of issue #418

vrqq:
The first problem in https://github.com/rbock/sqlpp11/issues/418
Match table name ```"public"."dk"``` to ```_literal[] = public.dk``` and ```struct PublicDK```


rbock:
Thanks for solving the first part!

I will be looking into the second part, but it might take a few days. If you're faster, I'd gladly merge again :-)

PR #421, opened 2022-02-11
add mediumtext data type for mysql in ddl2cpp script

linrongbin16:


rbock:
Note: github does send notifications for PRs. So usually, there is no need for a secondary ping on gitter :-)

linrongbin16:
OK got it

PR #427, opened 2022-03-11
Add errcode/sqlstate to postgres exception handling

MikeNeilson:
While using the library (with postgresql) I realized the result class wasn't sending the error code with the exception. This seems fine for the well named exception but not an database using user defined error codes.

I've created a "sql_user_error" class that stores the code and added a default case to the ThrowSQLError and added a test to verify operation.

In addition the ddl2cpp test wouldn't succeed when I ran `make test`. It appears that the code was changed recently and the cmake file wasn't updated. I took the liberty of adjusting the cmake file so the tests would execute.

If you're curious about the particular use case the API I've created inserted data into a view and a trigger handles the primary data validation. It's a time series database so several different errors that could happen like bad names, data off interval, etc and parsing the message text at the API level would be problematic.

### DIFF
--- a/include/sqlpp11/postgresql/exception.h
+++ b/include/sqlpp11/postgresql/exception.h
@@ -125,6 +125,30 @@ namespace sqlpp
       }
     };
 
+    /** Any error not covered by standard errors. 
+     * For example custom error code from a user
+     * defined pl/pgsql function
+     * 
+     */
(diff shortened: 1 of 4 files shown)


rbock:
Nice! This looks very good to me. Thanks for additional error class and the attention to detail in tests!

PR #438, opened 2022-04-17
Fix parsing of float fields with scale digits

jhunold:
After the last refactoring, ddl2cpp failed to parse float fields with scale specifiers like "NUMERIC(314,15)" in my SQL declarations. Fix with updated test data.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -198,7 +198,8 @@ ddlType = (
 )
 
 ddlUnsigned = pp.CaselessLiteral("UNSIGNED").setResultsName("isUnsigned")
-ddlWidth = ddlLeft + pp.Word(pp.nums) + ddlRight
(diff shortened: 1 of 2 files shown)


scripts/ddl2cpp
```diff
@@ -198,7 +198,8 @@ ddlType = (
 )
 
 ddlUnsigned = pp.CaselessLiteral("UNSIGNED").setResultsName("isUnsigned")
-ddlWidth = ddlLeft + pp.Word(pp.nums) + ddlRight
+ddlDigits = pp.Combine("," + pp.Word(pp.nums))
```
rbock, line 201: `ddlDigits = "," + pp.Word(pp.nums)`  should do the trick as well and allow for something like 

`NUMERIC(314, 15)`
(note the space after the comma)

rbock:
Thanks for digging into this!

PR #439, opened 2022-04-19
Added MariaDB keywords GENERATED and PERIOD

Carlitxxx86:
Both keywords are used for System Versioned tables.
GENERATED marks a column as auto generated.
PERIOD is ignored fro the script.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -217,6 +217,7 @@ ddlAutoKeywords = [
     "SMALLSERIAL",
     "SERIAL",
     "BIGSERIAL",
+    "GENERATED",
 ]
 ddlAutoValue = pp.Or(map(pp.CaselessLiteral, sorted(ddlAutoKeywords, reverse=True)))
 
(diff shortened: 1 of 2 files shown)


rbock:
Thanks for adding the keywords. Could you add a minimal example to [tests/scripts/ddl2cpp_sample_good.sql](https://github.com/rbock/sqlpp11/blob/main/tests/scripts/ddl2cpp_sample_good.sql) (e.g. as a new table)?

Double-check the results of `ctest -R sqlpp11.scripts.ddl2cpp.good_succeeds --verbose` in the generated `fail.h`.

Carlitxxx86:
Ok, already done. The test passes with warnings for using TIMESTAMP columns. 

36: Warning: date and time values are assumed to be without timezone.
36: Warning: If you are using types WITH timezones, your code has to deal with that.

Other than that, it seems ok.

Carlitxxx86:
Just as a note, when you define a table WITH SYSTEM VERSIONING, with implicit column definitions for ROW START and ROW END, the scripts generate a correct table definition, but without the columns for ROW START and ROW END, so there is no way for sqlpp to access them. If you want to access those columns you have to define them explicitly.

rbock:
Nice work, thanks.

As for the `WITH SYSTEM VERSIONING`: I´d be happy to accept a PR for supporting that (would require a slightly more involved test to make sure those two columns are actually created).

PR #440, opened 2022-04-23
Update ddl2cpp script to ignore OR REPLACE CLAUSE

Carlitxxx86:
ddl2cpp parser failed if the table was created with CREATE OR REPLACE TABLE... type of instructions. This patch ignores the OR REPLACE clause.


rbock:
Nice, I did not know the `OR REPLACE` clause :-)

Could you extend the test as well?

PR #445, opened 2022-06-06
Add support for "rationale" column type from pg_rational extension

jhunold:
Implement DDL2CPP support for the [pg_rational](https://github.com/begriffs/pg_rational) extension for PostgreSQL.
The generated column type is string as the C++ standard is still missing runtime ratio arithmetic.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -50,10 +50,12 @@ ddlOperator = pp.Or(
 
 ddlBracedExpression = pp.Forward()
 ddlFunctionCall = pp.Forward()
-ddlCast = ddlString + "::" + ddlTerm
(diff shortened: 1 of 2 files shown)


scripts/ddl2cpp
```diff
@@ -366,6 +369,17 @@ def testConstraint():
         assert result.isConstraint
 
 
+def testRational():
+    for tex in [
```
rbock, line 373: s/tex/text/

jhunold:
The test was completely broken. Thanks for spotting this. I've also added `ddl2cpp --test` to the auto-test suite to catch this in the future.

rbock:
Thanks for being so thorough!

PR #462, opened 2022-12-09
Fix some issues when compiling with MSVC

enwi:
- Fix `FindMariaDB.cmake` not finding include directories and use similar structure to `FindMySQL.cmake`
- Fix missing include for `iso646.h` when building for windows


rbock:
LGTM, thanks!

Issue #466, opened 2023-01-18
[ddl2cpp] Tables ignored when named by reserved keywords in postgresql (pg_dump)


fygonzalo:
Hi there. I started using the library, it's incredible.

I encountered a problem when trying to generate the cpp files when tables are named after a reserved keyword in postgresql (eg character, int, etc). When dumping with pg_dump, I noticed that those tables are dumped as:

```sql
CREATE TABLE public."character" (
    id integer NOT NULL
)
```

The problem is the ddl2cpp parser doesn't expect that expression in the table name and ignores it. It's trying to match `public."character"` to `ddlName`.

https://github.com/rbock/sqlpp11/blob/a77d8009bbca20d156bcfc4808a8c408712aabe6/scripts/ddl2cpp#L42-L46

- Doesn't match to `ddlTerm` because of the dot.
- Doesn't match to `ddlString` because the entire name isn't surrounded in quotes.
- Doesn't match to the `Combine` expression because only the second part is quoted.

I fixed it locally by adding `pp.Word(pp.alphas + "_", pp.alphanums + "_$")` as another option to the line 43, but I don't know much about the sql syntax and if this may break in other part.

Hope you can check this.

Thanks in advance!

rbock:
Hi,

Glad you like the library :-)

As for this issue, could you modify `ddlName` instead? Something like `pp.Combine(ddlTerm + "." + ddlString)` (untested)?
That way, the change would be specific to names, instead of modifying what a string looks like.

WDYT?

Thanks,
Roland

FYGonzalo:
You are right, it's more appropiate to put in `ddlName` instead. It should be `pp.Combine(ddlTerm + ddlString)` (without the dot) because `ddlTerm` already matches the dot.

It works for me and looks like won't cause any issue.

rbock:
Thanks!

PR #467, opened 2023-01-18
Extend match expression for table names in ddl2cpp

fygonzalo:
Using pg_dump, if the table is named after a special keyword (e.g. character), is dumped as public."character".
This commit extends the names expression to match those cases.


rbock:
Thanks for fixing #466

PR #470, opened 2023-01-30
ddl2cpp: allow inline column constraints ("PRIMARY KEY") but don't break auto-id

bloerwald:
The CREATE queries
```
CREATE TABLE tab (col INTEGER NOT NULL AUTO_INCREMENT PRIMARY KEY); -- mysql
CREATE TABLE tab (col INTEGER NOT NULL PRIMARY KEY AUTO_INCREMENT); -- mysql
CREATE TABLE tab (col INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT); -- sqlite, *no* AI before PK allowed!
```
mark the column `col` as both auto-data and primary key. ddl2cpp sadly only likes the first form, while the other order is also allowed by MySQL and **only** allowed by sqlite.

The queries parsed fine in v0.61, i.e. before the ddl2cpp rewrite, so I assume it was intended but missed in the rewrite.

The full syntax likely allows more than just the keywords, but that hopefully is suppressed by the existing `pp.Suppress(ddlExpression)`.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -227,21 +227,6 @@ ddlAutoKeywords = [
 ]
 ddlAutoValue = pp.Or(map(pp.CaselessLiteral, sorted(ddlAutoKeywords, reverse=True)))
 
-ddlColumn = pp.Group(
-    ddlName("name")
-    + ddlType("type")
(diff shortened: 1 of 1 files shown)


rbock:
Thanks for reporting and fixing! LGTM.

Issue #471, opened 2023-02-08
Float precision


bloerwald:
We are using a table containing floats, and have tests that use random values. The builtin serializers use `std::ostringstream` as underlying implementation, which defaults to `std::ios_base::precision() == 6`. This implies that values are quite quickly truncated.

A quick hack of 
```diff
diff --git a/include/sqlpp11/sqlite3/connection.h b/include/sqlpp11/sqlite3/connection.h
index 79b3ae6..bd1a8ea 100644
--- a/include/sqlpp11/sqlite3/connection.h
+++ b/include/sqlpp11/sqlite3/connection.h
@@ -159,4 +159,5 @@ namespace sqlpp
       serializer_t(const connection& db) : _db(db), _count(1)
       {
+          _os.precision (std::numeric_limits<long double>::digits10 + 1);
       }
```
shows that the underlying implementation supports higher precision values without issue.

Was this an explicit design choice or are float values just so rare nobody noticed?

rbock:
Hi, 

It was not a design choice. I believe that most folks avoid float values. I know that I do :-)

I am happy to accept a pull request to change the precision.

Best,
Roland

PR #472, opened 2023-02-09
#471 serializers: ensure float precision

bloerwald:
As described in #471, storing float values that need higher precision than the default `6` would be rounded. This was true for all but postgresql prepared statements. Those were only supporting up to `double` and were off by a bit (`maxdigits10` vs `digits10`).

Now, all streams I could find (except those in prepared statements for specific non-float types) should be using full precision. An internal test that rolls random `double` values worked fine with various high random values.

I'm not sure how to unit test this without having database backends, so a test is missing. Best I could come up with is serializing with some known-high-precision number and checking the string to be identical?


rbock:
floats are ~odd~ floats :-)

```
26/94 Test #26: sqlpp11.core.serialize.In .............................................Subprocess aborted***Exception:   0.11 sec
[...]/sqlpp11/tests/core/serialize/compare.h 52
Expected: -->|tab_foo.omega IN(1.7,2.5,17,0)|<--
Received: -->|tab_foo.omega IN(1.70000004768371582031,2.5,17,0)|<--
terminate called after throwing an instance of 'std::runtime_error'
  what():  unexpected serialization result
```

bloerwald:
Argh, sorry, I tried to reproduce it yesterday but managed to check out the wrong branch m(

I agree that one probably doesn't want it to serialise to that, although it of course *is* correct, just ugly, [even for numbers this low](https://godbolt.org/z/T6xKGjd5Y) (slightly adapted from cppreference). 

One can [set the precision for the float types specifically](https://godbolt.org/z/1bf6dEY9a) rather than using long-double for all. This does improve the ugliness slightly:  With 1.7 it still is `1.70000005` (float) `1.7` (double) and `1.69999999999999995559` (long double). 

The easiest fix for the failing test is using a safe float like `1.25` instead of `1.7`. The second easiest fix is using the ugly (but well defined) `1.70000004768371582031`. Since users can be confused by that value ([despite it being fine](https://godbolt.org/z/TzYTcTx8T)), the third option is using the specialised stream, which is adding a bit of overhead for visual effect only.

Do you have a preference?

rbock:
> One can [set the precision for the float types specifically](https://godbolt.org/z/1bf6dEY9a) rather than using long-double for all. This does improve the ugliness slightly: With 1.7 it still is `1.70000005` (float) `1.7` (double) and `1.69999999999999995559` (long double).

I like this one, with a serialize test for each to set expectations.

bloerwald:
> I like this one, with a serialize test for each to set expectations.

Done. The test has the boilerplate to use the specific serializers as well, but commented out since it requires the dependencies, and I didn't want to add all the "if dependency is found" infrastructure which I don't see existing yet. I can offer to instead add it to some random respective connector-specific test?

tests/core/serialize/Float.cpp
```diff
@@ -0,0 +1,122 @@
...
+  {
+    auto const value {sqlpp::value(10.0000114)};
+    auto const expected {"10.0000114"};
+    {
+      MockDb::_serializer_context_t serializer;
+      given_serializer_serializes_expr_as(__LINE__, serializer, value, expected);
+    }
+
+    // The following should also work, but require dependencies. The assumption is that
+    // since they use the wrapper it is fine.
```
rbock, line 98: "the wrapper" seems a bit vague?

tests/core/serialize/Float.cpp
```diff
@@ -0,0 +1,122 @@
...
+ * BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
+ * LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
+ * OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include <sqlpp11/data_types/floating_point.h>
+#include <sqlpp11/detail/float_safe_ostringstream.h>
+// #include <sqlpp11/mysql/connection.h>
```
rbock, line 28: These seem to be left overs?

Oh, I see. These are required in case someone uncomments the tests below?

Can you add serialize tests for the respective connectors, i.e.

tests/mysql/serialize/Float.cpp
etc

?

include/sqlpp11/detail/float_safe_ostringstream.h
```diff
@@ -0,0 +1,76 @@
...
+{
+  namespace detail
+  {
+    template<typename T, typename = void>
+    struct float_safe_ostringstream_implementation
+    {
+      template<typename U>
+      void operator() (std::ostringstream& os, U&& x) const
+      {
+        os << std::forward<U> (x);
```
rbock, line 48: nit: This space looks odd to me. Did you run this through clang-format?

include/sqlpp11/detail/remove_cvref.h
```diff
@@ -0,0 +1,64 @@
...
+#define SQLPP11_DETAIL_REMOVE_CVREF_H
+
+#include <type_traits>
+
+namespace sqlpp
+{
+  namespace detail
+  {
+    template<typename T>
+    using remove_reference_t = typename std::remove_reference<T>::type;
```
rbock, line 37: Do you have plans to use this? Otherwise, this whole header seems a bit overkill. `remove_cvref_t` is used in one location, everything else is not used at all?

rbock (changes requested):
Thanks for the update! This looks pretty thorough!
I added a couple of comments.
PTAL

include/sqlpp11/detail/float_safe_ostringstream.h
bloerwald: Yep, sorry, I tried to follow existing style but apparently failed in quite a few locations. Running through clang-format does change quite a lot of existing code, so I didn't think it was the current style. I'll run the files I touched through it and pick only those changes that also touch the lines I touched in the existing files.

include/sqlpp11/detail/remove_cvref.h
bloerwald: remove_cvref uses the other two, that's why i also added those. I assumed you used headers like this to emulate c++14+ features like enable_if_t. I'll inline it at the only use.

tests/core/serialize/Float.cpp
bloerwald: replaced with real type name

tests/core/serialize/Float.cpp
bloerwald: Done, sorry for all the boilerplate ;)

include/sqlpp11/detail/float_safe_ostringstream.h
rbock: OK, I should look into this then. The config is rather old by now and clang-format has been under heavy development :-)

include/sqlpp11/detail/remove_cvref.h
rbock: Ah, that's not what I meant. Sorry for being unclear:

I am all for `using` declarations and creating headers, but only if it makes a real difference. In this it is adding a lot of overhead just to be able to say `remove_cvref_t<T>` instead of `typename std::remove_cv<typename std::remove_reference<T>::type>::type` in a single place.

Note: There is also `std::decay` which should do the job, too?

include/sqlpp11/detail/float_safe_ostringstream.h
```diff
@@ -0,0 +1,87 @@
...
+      static_assert(std::is_same<remove_cvref_t<int&&>, int>{}, "");
+      static_assert(std::is_same<remove_cvref_t<int&>, int>{}, "");
+      static_assert(std::is_same<remove_cvref_t<int* const&&>, int*>{}, "");
+      static_assert(std::is_same<remove_cvref_t<int* const&>, int*>{}, "");
+      static_assert(std::is_same<remove_cvref_t<int>, int>{}, "");
+
+      template <typename T>
+      friend float_safe_ostringstream& operator<<(float_safe_ostringstream& os, T&& x)
+      {
+        float_safe_ostringstream_implementation<remove_cvref_t<T>>{}(os, x);
```
bloerwald, line 80:
```suggestion
      template <typename T>
      friend float_safe_ostringstream& operator<<(float_safe_ostringstream& os, T&& x)
      {
        float_safe_ostringstream_implementation<typename std::decay<T>::type>{}(os, x);
```
```suggestion
      template <typename T>
      using decay_t = typename std::decay<T>::type;
      
      template <typename T>
      friend float_safe_ostringstream& operator<<(float_safe_ostringstream& os, T&& x)
      {
        float_safe_ostringstream_implementation<decay_t<T>>{}(os, x);
```
```suggestion
      template <typename T>
      using remove_cvref_t = typename std::remove_cv<typename std::remove_reference<T>::type>::type;

      template <typename T>
      friend float_safe_ostringstream& operator<<(float_safe_ostringstream& os, T&& x)
      {
        float_safe_ostringstream_implementation<remove_cvref_t<T>>{}(os, x);
```
```suggestion
      template <typename T>
      friend float_safe_ostringstream& operator<<(float_safe_ostringstream& os, T&& x)
      {
        float_safe_ostringstream_implementation<typename std::remove_cv<typename std::remove_reference<T>::type>::type>{}(os, x);
```

include/sqlpp11/detail/remove_cvref.h
bloerwald: `decay` is a superset of `remove_cvref` and does things we don't need here (array, function pointer). It would work exactly the same though.

I added a comment with suggestions of the different variations, feel free to pick one of those :)

rbock: Cool, I was not aware of this suggestion feature. That's nice.

And thanks for playing through the options! Much appreciated!

rbock:
Thanks for going the extra mile! Looks very good to me at this point.

Please let me know if you are OK with merging at this point, too?

bloerwald:
> Please let me know if you are OK with merging at this point, too?

Seems fine to me. :+1: 

I had some very-close-to-zero values deserialize badly in our unit tests, but that seems to be a SQLite issue, not an sqlpp one:

```
input hex         float_safe_ostringstream  deserialized                  output hex
80a797b1d9baf781  -1.6798458567329104e-305  -1.67984585673291039419e-305  80a797b1d9baf781
80a797b1d9baf782  -1.6798458567329106e-305  -1.67984585673291064715e-305  80a797b1d9baf782
```
Serializer is fine :arrow_up: SQLite truncates.  :arrow_down: 
```
sqlite> CREATE TABLE test (value REAL);
sqlite> INSERT INTO test VALUES (-1.6798458567329104e-305);
sqlite> INSERT INTO test VALUES (-1.6798458567329106e-305);
sqlite> SELECT * FROM test;
-1.67984585673291e-305
-1.67984585673291e-305
```
I think I'll change my tests to not assert bit-identical results ;)

rbock:
Done, thanks!

> Edit: allowed for some inaccuracy, got 5000 successful test runs.

:+1:

PR #473, opened 2023-02-10
ci: enable ctest output on failure

bloerwald:


rbock:
Good thinking, thanks!

PR #478, opened 2023-04-02
Add possibility to assign a std::string_view to a prepared statement parameter

matthijs:
Make it possible to assign a std::string_view to a prepared statement.

See as example (I've added it also to the `Insert.cpp`):
```c++
auto prepared_insert_sv = db.prepare(insert_into(t).set(t.gamma = parameter(t.gamma), t.delta = parameter(t.delta), t.beta = parameter(t.beta)));
prepared_insert_sv.params.gamma = true;
prepared_insert_sv.params.delta = 17;
std::string_view sv = "string_view";
prepared_insert_sv.params.beta = sv;
db(prepared_insert_sv);
```


rbock:
Nice, thanks!

Issue #481, opened 2023-05-22
Inserting std::chrono::system_clock::time_point into timestamp with time zone


MeanSquaredError:
Currently when sqlpp11 inserts/updates a timestamp field it generates timestamp literals without a time zone. When such literals are inserted into a field that is a timestamp with time zone, PostgreSQL interprets the generated timestamp literals as having the local time zone, which causes problems.

Let's say that we have the number 1684706721, which is the number of seconds since the start of the UNIX epoch (1970-01-01 00:00:00 UTC), i.e. 2023-05-21 22:05:21 UTC.

This is a simplified version of the code that generates the insert:
```
int myts = 1684706721;
sqlpp::postgresql::insert_into(tgm).set (tgm.message_ts = std::chrono::system_clock::time_point {std::chrono::seconds {myts}})
```

When inserting the timestamp sqlpp11 generates the following SQL query
```
INSERT INTO messages (message_ts) VALUES(TIMESTAMP '2023-05-21 22:05:21.000000000')
```

The above SQL code inserts a timestamp that uses the local time zone (+03). A select of the inserted value returns `2023-05-21 22:05:21+03`, which is obviously the wrong value.

I think a possible workaround would be to change sqlpp11 so that it would generate a timestamp in the +00 timezone. In this case it would be 
```
INSERT INTO messages (message_ts) VALUES(TIMESTAMP WITH TIME ZONE '2023-05-21 22:05:21.000000000+00')
```
This will tell PostgreSQL that we insert a timestamp in the UTC timezone, so the correct value will be inserted.

rbock:
Thanks for the detailed comment. That sounds plausible.

I'd be happy to accept a pull request with this change and unit tests as confirmation.

MeanSquaredError:
I have been using a modified version of sqlpp11 (as per my suggestion above) for about a couple of weeks. The modified version works correctly with PostgreSQL + timestamp with time zone.

The change is available here:
https://github.com/MeanSquaredError/sqlpp11/commit/cc24cfb21da5e6a5e1849dab54ab75099ae873cd

In the next couple of days I will add tests and then will provide a PR.

rbock:
Great! Looking forward to the PR :-)

PR #487, opened 2023-06-13
Treat PostgreSQL time values as being in UTC time zone

MeanSquaredError:
This PR fixes the problem that is mentioned here
https://github.com/rbock/sqlpp11/issues/481

Currently when PostgreSQL timestamptz and timetz fields are read from/written they are treated as being in the local timezone and there is no easy way to write values in the UTC timezone. 

This is a problem because PostgreSQL timestamptz and timetz fields are UNIX timestamps (they only change their output representation based on the configured server timezone) and std::chrono::time_point values are UNIX timestamps (until C++20 it was an implementation detail and starting from C++20 it is specified explicitly in the standard). At the same time writing a time_point with sqlpp11 and then reading the saved UNIX timestamp (e.g. with extract(epoch from field_name) ) yields a different value.

Additionally mapping from the local timezone to the UTC timezone is not always unique, so it makes sense to use UTC whenever possible and only convert from UTC to the local time zone when necessary.

Ideally for C++20 we could support reading/writing std::chrono::zoned_time and std::chrono::local_time which would allow one to choose the timezone explicitly, but for C++11 it makes sense just to default to UTC instead of the local time zone.

### DIFF
--- a/tests/postgresql/usage/TimeZone.cpp
+++ b/tests/postgresql/usage/TimeZone.cpp
@@ -1,0 +1,159 @@
+/**
+ * Copyright © 2023 Vesselin Atanasov
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ *   Redistributions in binary form must reproduce the above copyright notice,
+ *   this list of conditions and the following disclaimer in the documentation
+ *   and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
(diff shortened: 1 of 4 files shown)


rbock:
Thanks for the PR! Looks good at first glance, including the much shorter code using regex.

I'll take a closer look, soon.

tests/postgresql/usage/TimeZone.cpp
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  void check_saved_values(sqlpp::postgresql::connection& dbc, sqlpp::chrono::microsecond_point tp, std::chrono::microseconds tod, sqlpp::chrono::day_point dp)
+  {
+    model::TabDateTime tab {};
+
+    // Check if the date/time values are saved using the UTC time zone
+    const auto &rows_1 = dbc(
+      select(
+        sqlpp::verbatim<sqlpp::integer>("floor(extract(epoch from c_timepoint)*1000000)::int8").as(sqlpp::alias::a),
```
rbock, line 90: I admit, I have a hard time understanding what is selected here. Can you add a comment?

rbock (commented):
Thanks again for the PR!
The library code looks good to me, but I am struggling a bit with the test, see comment.

Best,
Roland

tests/postgresql/usage/TimeZone.cpp
MeanSquaredError: Indeed the test seems to be more cryptic than the main code :-)
I added a few extra comments that hopefully make the test code clearer.
So please let me know if it needs additional clarification.

rbock: I think this is clear enough now. Thanks for adding the comments.

rbock:
Thanks again for the PR! I always try to avoid timezones, which is why my experience in that area is quite limited :-)

MeanSquaredError:
Thank you once again for the really useful library!

From my experience in 99% of the cases the best approach is to keep timestamps as a (micro)seconds offset from the beginning of the UNIX epoch (1970-01-01 00:00:00 UTC).

Internally PostgreSQL takes a similar approach and their timestamp with time zone is just a microseconds offset  from the beginning of the UNIX epoch. Ironically their "timestamp with time zone" doesn't really have its own time zone - it is just a global (per-server) timezone which is used by the server to parse the input string to a microseconds integer and format the microseconds integer to an output string.

Anyway, I guess in the future it is possible to add support in sqlpp for std::chrono::zoned_time (for those who want to specify the time zone explicitly) and std::chrono::local_time (if someone really wants to work with time in the local timezone).

rbock:
Indeed, `zoned_time` looks like a very good match.

Issue #485, opened 2023-06-08
Discussion: Custom Types


CJCombrink:
This is not an issue (yet) but more a question around custom types.

PostgreSQL (which we use) supports creating new types, using [CREATE DOMAIN](https://www.postgresql.org/docs/current/sql-createdomain.html)
Example:
```
CREATE DOMAIN FREQUENCY AS BIGINT;
```

How would be the best to use this in our sql schema, along with sqlpp11?
I am aware of the message stating we should update the script manually, but that is not ideal since I want to be able to use the library without "hacking" the script (we are using the library through a package manager, conan specifically).

In my mind I can think of the following options:
1. Update the ddl2ccp script to understand `CREATE DOMAIN` to extend/update the type conversion lists
   - This might be too PostgreSQL specific and I can't comment about such support in other databases.
   - The following comment in the linked page might make this less of a problem: "The command CREATE DOMAIN conforms to the SQL standard."
2. Add an extension point to the dll2cpp through a command line argument:
   - Suggestion: `--type-mapping-file:     Path to a text file that adds custom type mappings 'extendedDdlIntegerTypes = [FREQUENCY]'`
   - This file is then loaded on startup of the script and the internal types are extended to support these.
   - This allows us to keep the mappings separate without touching the script
 3. Fall back to text and still generate the files
    - My guess is many types can be passed to the database as text, which is not ideal specifically for integer types like my example.

Just supporting `CREATE DOMAIN` does however not cover all uses cases:
1. We use the UUID type which is not our type, but not supported by the script. 
2. Extensions can also add types 
3. We also create our own types which does not fit in with option 1 ([CREATE TYPE](https://www.postgresql.org/docs/current/sql-createtype.html) and supporting that as well can be overly complex.
    - For this specific example we will probably remove the usage of the custom type or use it as a string

Option 2 would be my hope. 1 can be nice addition.

rbock:
Hi Carel,

Thanks for the detailed comments. I was not even aware of `CREATE DOMAIN` :-)

I like the second option2, too. 

In addition to what you wrote: 
The extension should probably issue a warning, if it would overwrite an "existing" type. This could happen, for instance, if we added support for `UUID` to ddl2cpp.

Would you be interested in giving it a try to write such an extension point?

Cheers,
Roland

PR #490, opened 2023-06-14
Updated the ddl2cpp script to allow custom/extended types

CJCombrink:
…ernal script

- Had to re-order the script to allow the command line to be parsed before setting up the parser
- Added an example python file to show usage
- Fixed tests as well

Implements #485

### DIFF
--- a/datatypes.py
+++ b/datatypes.py
@@ -1,0 +1,16 @@
+# Custom datatype mappings
+# Maps datatypes unknown to ddl2cpp to known types
+# This file can be copied, updated and then passed to ddl2cpp as a command line argument
+# Each variable in this file must be a list of case insensitive type names.
+# Usage:
+#    ddl2cpp [other optional args] --datatype-file=/path/to/datatypes.py <path to ddl> <path to target> <namespace>
+
+extendedDdlBooleanTypes = []
+extendedDdlIntegerTypes = []
+extendedDdlSerialTypes = []
+extendedDdlFloatingPointTypes = []
+extendedDdlTextTypes = ["UUID", "other_type"]
+extendedDdlBlobTypes = []
+extendedDdlDateTypes = []
+extendedDdlDateTimeTypes = []
+extendedDdlTimeTypes = []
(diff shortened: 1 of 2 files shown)


scripts/ddl2cpp
```diff
@@ -485,6 +494,19 @@ def setArgumentBool(s, bool_value):
     var_name = first_lower(re.sub("(\s|-|[0-9])(\S)", repl_func_for_args, s))
     globals()[var_name] = bool_value
 
+def loadExtendedTypesFile(filename):
```
CJCombrink, line 497: One can probably make an argument for not loading a python script but a csv or other text file. I am not very proficient with python and decided with going this route. It should be easy enough to update/extend at this function without having to touch the rest of the code again.

rbock: Could you try to go the CSV route? Reading CSV is quite simple in python.

rbock (changes requested):
Thanks for creating this PR!

The commit message starts with `...`? That looks a bit like an accident?

Did you forget to add the tests you mentioned? There should be be at least one test using the new command line switch.

scripts/ddl2cpp
CJCombrink: Sure I can give it a bash and see what I get

CJCombrink:
> Thanks for creating this PR!
> 
> The commit message starts with `...`? That looks a bit like an accident?
> 
> Did you forget to add the tests you mentioned? There should be be at least one test using the new command line switch.

datatypes.py
```diff
@@ -0,0 +1,16 @@
...
+# This file can be copied, updated and then passed to ddl2cpp as a command line argument
+# Each variable in this file must be a list of case insensitive type names.
+# Usage:
+#    ddl2cpp [other optional args] --datatype-file=/path/to/datatypes.py <path to ddl> <path to target> <namespace>
+
+extendedDdlBooleanTypes = []
+extendedDdlIntegerTypes = []
+extendedDdlSerialTypes = []
+extendedDdlFloatingPointTypes = []
+extendedDdlTextTypes = ["UUID", "other_type"]
```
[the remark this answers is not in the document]
CJCombrink: Was thinking more of an example

scripts/ddl2cpp
CJCombrink: Do you have an idea or suggestion on how a CSV would look for this?
Maybe:
```
extendedDdlIntegerTypes, "one", "two"
```
I think this is what tripped me up when I considered the approach, I could not think of a nice other format

CJCombrink: I think I have something that can work, will create a new PR once it works

PR #494, opened 2023-06-19
Const connection config

CJCombrink:
- The postgres connection does not change the connection_config thus passing it as const.
- Implies 'thread safety' when using the same config for multiple connections


rbock:
Looks good to me, thanks for the cleanup!

PR #496, opened 2023-06-27
gcc warnings

MeanSquaredError:
gcc issues several warnings when compiling the current codebase. After investigating the warnings it turned out that they point to a couple of real issues:

1. The constructors of plpgsql_no_data_found try to std::move() objects that are const std::string&
Obviously moving const references does not work, and that is why gcc complains about "redundant moves". The PR changes the constructor parameters to std::string just like the the constructor parameters of the other PostgreSQL errors.

2. There are a bunch of dangling references to fetched SQL rows in tests/core/types/result_row.cpp
The PR fixes these too.

### DIFF
--- a/include/sqlpp11/postgresql/exception.h
+++ b/include/sqlpp11/postgresql/exception.h
@@ -125,21 +125,21 @@ namespace sqlpp
       }
     };
 
-    /** Any error not covered by standard errors. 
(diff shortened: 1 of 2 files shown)


rbock:
Thanks for the cleanup!

PR #499, opened 2023-07-06
Connection pools

MeanSquaredError:
This PR adds support for connection pooling to the PostgreSQL, MySQL and SQLite3 connectors. It also adds tests for the new connection pooling code and updates the sample connector API classes to match the new updated structure of the database connectors.

The connection pooling code is based on the code from sqlpp17, but it also has a number of changes that make it different from the sqlpp17 implementation.

Below is a list of the main differences from a user perspective:

- The sqlpp11 connection pools accept a capacity parameter just like the sqlpp17 version, but this capacity can be increased dynamically if the initial capacity turns out to be insufficient. The sqlpp17 version just drops the old connections and replaces them by new ones.

- The sqlpp11 pooled connections can be moved without losing of connection handles or insertion into the pool of invalid (empty) handles.

- Due to the unification of connector implementations all non-pooled connectors now can be created either from "const std::shared_ptr<const connection_config>&" or from "const connection_config&". Pooled connections cannot be created by the user, they are only created by the connection pool. Also all non-pooled connections now support connectUsing() (originally only the PostgreSQL connector had it).

There are also many changes in the internals. Below are the most important ones:

- The sqlpp17 connection pools had separate implementations for the different database engines. The pools in sqlpp11 use a common template class.

- The sqlpp17 connections were templates having the pool class as a parameter. The base connecitons classes in sqlpp11 are not templates. The non-pooled and pooled classes are templates that derive from the base connection classes. This allows to move most of the common code in the conn_normal and conn_pooled templates, which arguably makes the implementation of the connection pooling code simpler.

- The format connection_handle classes has been unified to a certain extent and I tried to document that in the sample connector api class.

- The connection handles that are stored in the base connector classes are always stored as std::unique_ptr<connection_handle>, because these unique_ptr objects are stored by the connection pool and are used to create connections when requested.

- I also tried to cleanup a bit the code of the MySQL tests.

Admittedly it is a big PR with lots of changes, so please let me know what you think about it.

### DIFF
--- a/connector_api/connection.h
+++ b/connector_api/connection.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (c) 2013-2015, Roland Bock
+ * Copyright (c) 2023, Vesselin Atanasov
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without modification,
@@ -37,6 +38,44 @@ namespace sqlpp
 {
   namespace database
   {
+    // Connection configuration that is used to create new database connections
+    struct connection_config
+    {
+      // Put the configuration properties here, e.g.
+      //
+      // std::string host;
+      // unsigned port;
+      // std::string username;
+      // std::string password;
+      // std::string db_name;
+    };
+
+    // The connection handle is a low-level representation of a database connection.
+    // Pre-connected handles are stored in the connection pool and are used to create
+    // full connection objects on request.
+    class connection_handle
+    {
+    public:
+      // Connection handles can be created from connection configurations
+      connection_handle(const std::shared_ptr<const connection_config>& config);
+
+      // Connection handles cannot be copied
+      connection_handle(const connection_handle&) = delete;
+      connection_handle& operator=(const connection_handle&) = delete;
+
+      // Connection handles can be moved
+      connection_handle(connection_handle&&);
+      connection_handle& operator=(connection_handle&&);
+
+      // Used by the connection pool to check if the connection handle is still
+      // connected to the database server
+      bool check_connection();
+
+      // Optional method that returns a native (low-level) database handle.
+      // Used by the test code to test the connection pool
+      native_db_handle native_handle();
+    };
+
     // The context is not a requirement, but if the database requires
     // any deviations from the SQL standard, you should use your own
     // context in order to specialize the behaviour, see also interpreter.h
@@ -48,9 +87,25 @@ namespace sqlpp
       std::string escape(std::string arg);
     };
 
-    class connection : public sqlpp::connection  // this inheritance helps with ADL for dynamic_select, for instance
+    // The base database-specific connection class. Non-pooled and pooled connection classes derive from it
+    class conn_base : public sqlpp::connection  // this inheritance helps with ADL for dynamic_select, for instance
     {
     public:
+      // Base configuration
+      using _conn_base_t = conn_base;
+
+      // Type of configuration instances
+      using _config_t = connection_config;
+
+      // Shared pointer wrapping a configuration instance
+      using _config_ptr_t = std::shared_ptr<const _config_t>;
+
+      // Type of connection handles
+      using _handle_t = connection_handle;
+
+      // Unique pointer wrapping a connection handle
+      using _handle_ptr_t = std::unique_ptr<_handle_t>;
+
       using _traits = ::sqlpp::make_traits<
           ::sqlpp::no_value_t,
           ::sqlpp::tag::enforce_null_result_treatment  // If that is what you really want, leave it out otherwise
@@ -62,13 +117,6 @@ namespace sqlpp
           ;
       // serializer and interpreter are typically the same for string based connectors
       // the types are required for dynamic statement components, see sqlpp11/interpretable.h
-
-      connection(...);
-      ~connection();
-      connection(const connection&) = delete;
-      connection(connection&&) = delete;
-      connection& operator=(const connection&) = delete;
-      connection& operator=(connection&&) = delete;
 
       //! "direct" select
       template <typename Select>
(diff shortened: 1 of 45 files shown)


MeanSquaredError:
I just added an update which has the following changes:
- The previous implementation of connection pools in sqlpp11 (and sqlpp17) had a bug which manifested when the connection pool was destroyed before all of the connection that it had given out were destroyed. In that case upon destruction a connection tried to return its handle to the pool (which was already destroyed) causing a crash. This update fixes the problem by creating a separate "connection pool core" object which handles giving out/returning of connections. The core is kept alive even if the main connection pool is destroyed. The core is destroyed after all connections to it are destroyed (essentially it is an object wrapped in std::shared_ptr).
- Additionally the public methods of connection_pool are extended and connection pools now can be default-constructed, moved. The default-constructed pools can be initialized afterwards via a new initialize() method.

MeanSquaredError:
Added a test for the case when a connection pool is destroyed before all of the handles that it has handed out.

connector_api/connection.h
```diff
@@ -48,9 +87,25 @@ namespace sqlpp
       std::string escape(std::string arg);
     };
 
-    class connection : public sqlpp::connection  // this inheritance helps with ADL for dynamic_select, for instance
+    // The base database-specific connection class. Non-pooled and pooled connection classes derive from it
+    class conn_base : public sqlpp::connection  // this inheritance helps with ADL for dynamic_select, for instance
```
rbock, line 91: Can you rename `conn` to `connection`?

MeanSquaredError: Sure. Should I rename all conn_xxxxx classes to connection_xxxxx, i.e. the ones in :
sqlpp11/include/sqlpp11/connection.h
sqlpp11/include/sqlpp11/mysql/connection.h
sqlpp11/include/sqlpp11/postgresql/connection.h
sqlpp11/include/sqlpp11/sqlite3/connection.h

MeanSquaredError: OK, I renamed:
conn_base -> connection_base
conn_normal -> connection_normal
conn_pooled -> connection_pooled
_conn_base_t -> _connection_base_t
_conn_pooled_t -> _connection_pooled_t
ConnBase -> ConnectionBase

The tests pass normally, the only (small) issue is that now it is more difficult to tell `connection_pooled` from `connection_pool`

rbock: Thanks!

> the only (small) issue is that now it is more difficult to tell `connection_pooled` from `connection_pool`

We could rename `connection_pooled` to `pooled_connection`. That would be more natural anyways.

WDYT?

MeanSquaredError: I think it is OK, but then we should probably rename `connection_normal` to `normal_connection` too?

rbock: That sounds good to me, yes.

MeanSquaredError: OK, I renamed

connection_pooled -> pooled_connection
connection_normal -> normal_connection
_connection_pooled_t -> _pooled_connection_t

tested the PostgreSQL, MySQL and SQLite3 connectors and pushed the update.

include/sqlpp11/sqlite3/connection.h
```diff
@@ -589,12 +512,110 @@ namespace sqlpp
...
+      _handle_ptr_t _handle;
+
+      // Constructors
+      connection_base() = default;
+      connection_base(_handle_ptr_t&& handle) : _handle{std::move(handle)}
+      {
+      }
+
+    private:
+      enum class transaction_status_type
```
rbock, line 526: Is there a technical reason for moving this code from above?

If no, then I would prefer to move the code in a separate PR.

connector_api/connection.h
rbock: Great, thanks!

rbock (commented):
This looks very good to me, with a question about moved code.

include/sqlpp11/sqlite3/connection.h
MeanSquaredError: No real technical reason, so I moved the private part of the class back to the top. I also ran the SQLite3 connector tests and committed the update.

My rationale for moving the private part to the bottom was as follows:

- At the top we have some public types, like _config_t, _config_ptr_t, _handle_t, etc. Not all of them should be public but I left them public for convenience (mostly for simpler testing code).
- After we have defined the public types, we can use them to define the handle member `_handle_ptr_t _handle`. The handle object needs to use the public type _handle_ptr_t and it also should stay protected (no need to be public, but it should be accessible from the derived class pooled_connection, so the only choice is protected).
- After we have defined the _handle member we can use it in our private code. Technically our private code can use the _handle member even before it is defined, but arguably it the code is more readable if the member data is defined before the code that uses it.

So that is why I ended with following order 
public: -> protected: -> private:

Overall it is just a matter of preferences, so  I moved the private code back to the top to simplify the review of the code.

rbock: Thanks! Can you do the same for the mysql connector as well (sorry for not explicitly asking for that before)?
This is a massive change. Reducing the size is important.

I am happy to discuss moving the code as you suggested in a second step :-)

MeanSquaredError: OK, I reordered back the code in the MySQL and PostgreSQL connectors too and also reverted a few code cleanups in the PostgreSQL connector in order to reduce the size of the diff.

rbock: Great, thanks!

rbock:
At this point, I am ready to merge. Please let me know whether your're OK with merging, too.

MeanSquaredError:
I am OK with the merge too, so please add the PR to the codebase.

There are several other improvements/fixes that can be made, but as you said these can go into separate PRs.

rbock:
Done :-)

Looking forward to future PRs!

PR #504, opened 2023-07-22
Fix serialization of PostgreSQL time data types

MeanSquaredError:
I changed slightly the PostgreSQL TimeZone test and made it setup a database session in the UTC+1 timezone instead of the default UTC timezone. This allowed it to find a few bugs that the UTC timezone was masking. So I prepared a PR with the fixes. More specifically this PR has the following changes:

- As mentioned above the TimeZone test is changed to use UTC+1 during the test.
- The serializaton of `time_of_day_operand` has been fixed. The previous PostgreSQL PR changed the parsing of "TIME WITH TIMEZONE" results so that their timezone was taken into an account, but did not fix the serialization. This PR fixes that. (The serialization of normal (non-prepared) time_point was fixed by the first PostgreSQL PR).
- The serialization of `time_point` for prepared statements has been fixed.
- The serialization of `time_of_day_operand` for prepared statements has been fixed.

This PR conflicts slightly with the other PR that fixes MySQL time handling. So once the MySQL PR is merged I will rebase this PR against the updated codebase.

### DIFF
--- a/include/sqlpp11/postgresql/prepared_statement.h
+++ b/include/sqlpp11/postgresql/prepared_statement.h
@@ -199,10 +199,9 @@ namespace sqlpp
       {
         const auto time = ::date::make_time(*value) ;
 
-        // Timezone handling - always treat the value as UTC.
-        // It is assumed that the database timezone is set to UTC, too.
+        // Timezone handling - always treat the local value as UTC.
         std::ostringstream os;
-        os << time;
+        os << time << "+00";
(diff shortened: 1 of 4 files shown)


MeanSquaredError:
@rbock 
I rebased this PR against the latest main branch and tested the core+the three connectors (all tests passed correctly).

rbock:
Thanks, looks good to me!

PR #505, opened 2023-07-24
Fix typo.

orangecheers-x:
change "MariaBD" to  "MariaDB".


rbock:
Thanks for spotting and fixing!

Issue #500, opened 2023-07-17
Future improvements of sqlpp11


MeanSquaredError:
@rbock Thank you for merging the connection pools into the main branch.

I made a list of things that may be improved in future versions of sqlpp.

**BUGS:**

- [DONE] ~~One of the MySQL tests (sqlpp11.mysql.usage.DateTime) fails on my development machine. This is not related to the new connection pools code, because it failed before too. I am not sure of the exact reason, one (unlikely) reason could  be because I am using the MySQL connector to connect to a MariaDB. I will try investigating and fixing it.~~
- [DONE] ~~The MariaDB and SQLCipher connectors fail to build on my machine. I took a quick look and it seems that the cmake for these two has problems. I will investigate that in more detail.~~

**IMPROVEMENTS OF THE CURRENT FUNCTIONALITY**
- [DONE] ~~Currently the connectors have different non-standard ways to check if the connection/handle is connected to the server. MySQL has a connection::is_valid() which essentially calls mysql_ping(). PostgreSQL just calls PQstatus() which checks passively if the remote server has closed the connection or if a previous request had an error. SQLite has nothing :-) 
Currently when the connection pool checks the handle status it just uses whatever functionality the connector supports which varies depending of the connector.
So it really makes sense to standardize this. Each connection and connection_handle should have two methods:~~
~~1. is_connected() checks if the connection has been established and the remote side has not closed the connection.~~
~~2. ping_server() actively pings the server. For PostgreSQL and SQLite3 we can just use the approach of PHP's pg_ping() that sends `SELECT 1` to the server and checks if there is an error.~~

~~Once we have standardized the is_connected() and ping_server() for connections, we can enhance slightly connection_pool::get() and make it accept a parameter which defines the type of the connection check ,e.g:~~
```
enum CHECK_TYPE
{
     CHECK_NONE,
     CHECK_CONNECTED,
     CHECK_PING
};

connection_type connection_pool::get (CHECK_TYPE check = CHECK_CONNECTED);
```
~~Then based on the requested check type it will perform the corresponding check before giving out the handle. Currently the MySQL connection pool performs a server ping before giving out a handle, which may be unnecessary if the MySQL server is reliable enough. So it makes sense to allow the user to choose the check type.~~
- Switching the testing framework to Catch2. sqlpp17 uses it, so it probably makes sense to use it in sqlpp11 too.

**CODE CLEANUPS** 
These are a matter of taste, but it still probably makes sense to discuss and implement some/all of them.

[...]

PR #506, opened 2023-07-24
Replace include guards with "#pragma once"

MeanSquaredError:
As per our previous discussion here https://github.com/rbock/sqlpp11/issues/500
this PR replaces the header include guards with `#pragma once` in all header files and in the python generator scripts. All tests (core+postgresql+mysql+sqlite3) ran successfully.

The `#pragma once` is not standard but is supported by all major C++ compilers and easier to maintain than the old-style #ifndef #define...#endif include guards.


MeanSquaredError:
I did automate the change, but then had to go manually over all the files just to make sure that the regular expressions in my script updated the files correctly :-)

PR #508, opened 2023-07-27
Fix building with sqlcipher

MeanSquaredError:
This PR backports some CMake code from sqlpp17, which fixes a build issue with SQLCipher.

It seems that CMake does not have builtin support for finding the SQLCipher library and when the sqlpp11's CMakeLists.txt tries to run `find_package(SQLCipher REQUIRED)`, it fails. 

This PR makes the following changes:
1. It moves the custom CMake scripts that sqlpp11 uses to find different libraries into their own subdirectory:
`cmake/Find*.cmake -> cmake/modules/Find*.cmake`
(I copied this layout from sqlpp17 as I think it makes more sense)
2. It copies the script `FindSQLCipher.cmake` from sqlpp17 into cmake/modules/, which solves the problem with the finding of the SQLCipher library.

This PR was built with
```
cmake -B build -DBUILD_POSTGRESQL_CONNECTOR=ON -DBUILD_SQLITE3_CONNECTOR=ON -DBUILD_MYSQL_CONNECTOR=ON -DBUILD_SQLCIPHER_CONNECTOR=ON -DBUILD_TESTING=ON -DUSE_SYSTEM_DATE=ON -DDEPENDENCY_CHECK=ON
cmake --build build -- VERBOSE=1
```
After building all tests passes successfully


MeanSquaredError:
PR update: Fixed copying of scripts from `cmake/modules/Find*.cmake ` during installtion.

rbock:
Thanks! Looks good to me (albeit not being knowledgeable in CMake).

MeanSquaredError:
Well, I am not a big fan of CMake either, because I think that an old-style Makefile is just as manageable as CMakeLists.txt.

However, nowadays everyone and his dog are using CMake so there is not much choice :-)

PR #509, opened 2023-08-01
Cleanup the code of the MySQL connector

MeanSquaredError:
OK, I updated the PR based on our discussion and I think it is ready for a review.

~~I mentioned my concern regarding the renaming of the types ending in _t here
https://github.com/rbock/sqlpp11/issues/500#issuecomment-1661392856
So it probably makes sense to postpone the merging of this PR until it is clear what to do with the renaming of these types.~~

This PR cleans up the code of the MySQL connector as per our discussion here
https://github.com/rbock/sqlpp11/issues/500#issuecomment-1659571760

The PR makes the following changes to the MySQL connector:

1. The code for the classes `connection_handle`, `prepared_statement_handle` and `result_handle.h` are moved to separate files in the 
`include/sqlpp11/mysql/detail/` directory.
~~2. All class and struct type names ending in `_t` have that suffix removed from their names. In some cases this caused some local variables and/or function parameters to be renamed in order avoid confusion and/or name clashes.~~
2. `sqlpp::mysql::connection_handle_t` has been renamed to `sqlpp::mysql::connection_handle` as the other two connectors (PostgreSQL and SQLite3) have `connection_handle` without the `_t` suffix.
3. `sqlpp::mysql::serializer_t` has been renamed to `sqlpp::mysql::context_t`
4. Const qualifier has been added to `sqlpp::mysql::context::escape()` and its parameter has been changed from `std::string` to `const std::string&`.
~~5. Old-style assignment-style parentheses-style initialization has been replaced with brace-style initialization. The only place where assignment-style initialization remains is `auto var = {value}` where C++11 has problems.~~
6. There were several variables and one helper class using CamelCase. Those were renamed to snake_case as the project as a whole is using snake_case.

After making the changes the project was built with tests for the MySQL, PostgreSQL and SQLite3 connectors and all tests passed successfully.

### DIFF
--- a/include/sqlpp11/mysql/bind_result.h
+++ b/include/sqlpp11/mysql/bind_result.h
@@ -26,91 +26,21 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#include <iostream>
-#include <memory>
-#include <vector>
 #include <sqlpp11/chrono.h>
 #include <sqlpp11/exception.h>
+#include <sqlpp11/mysql/detail/prepared_statement_handle.h>
 #include <sqlpp11/mysql/sqlpp_mysql.h>
 
+#include <iostream>
 #ifdef _MSC_VER
 #include <iso646.h>
 #endif
+#include <memory>
 
 namespace sqlpp
 {
   namespace mysql
   {
-    namespace detail
-    {
-      struct result_meta_data_t
-      {
-        size_t index;
-        unsigned long bound_len;
-        my_bool bound_is_null;
-        my_bool bound_error;
-        std::vector<char> bound_text_buffer;  // also for blobs
-        const char** text_buffer;
-        size_t* len;
-        bool* is_null;
-      };
-
-      struct prepared_statement_handle_t
-      {
-        struct wrapped_bool
-        {
-          my_bool value;
-
-          wrapped_bool() : value(false)
-          {
-          }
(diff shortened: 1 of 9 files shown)


rbock:
Thanks for experimenting with this. I think it makes sense to move code as suggested in the first three commits. I would prefer them as individual commits (if that makes sense, I have not checked them in detail).

The renaming requires more discussion, I think.

MeanSquaredError:
@rbock 
OK, I made changes to the PR and left the `_t` suffixes unchanged except for `connection_handle`. The other two connectors have `connection_handle` without the `_t` suffix, so I removed it from the MySQL connection handle too.

I think the PR is ready for a review, so please take a look.

rbock:
Looks good to me.

My only concern: This mixes several topics. In case something were broken with this CL, it would be easier to figure out the cause if they could be separate. The biggest change that is not related to moving code is the use of brace style initializers.

If this can be separated into its own PR easily, I would prefer that.

MeanSquaredError:
@rbock 
OK, I removed the braced-init commit from the PR and tested it - all tests passed.
I will submit the braced-init changes in a separate PR.

Please review this PR and let me know if further changes are required.

MeanSquaredError:
Force-pushed an update that removed a braced init that sneaked into this PR from the other PR.

I think now this PR is ready for a review.

rbock:
Thanks for separating the changes! This looks good to me.

PR #510, opened 2023-08-05
Use brace-style initialization in the MySQL connector code

MeanSquaredError:
This is the second PR which cleans up the MySQL connector code. This one replaces old-style variable initializations with {}-initializations.
The only places where the old-style initialization has been kept is
- `auto var={value}` where C++11 uses an initializer_list.
- `std::vector(size_type count, ...)` which cannot be invoked via the {} syntax.

In the PostgreSQL connector we also have one case of
- `std::string(size_type count, ...)` which cannot be invoked via the {} syntax. However in the MySQL connector we don't have any uses of this constructor.

This PR was built and tested with
```
```

### DIFF
--- a/include/sqlpp11/mysql/bind_result.h
+++ b/include/sqlpp11/mysql/bind_result.h
@@ -44,11 +44,11 @@ namespace sqlpp
     class bind_result_t
     {
       std::shared_ptr<detail::prepared_statement_handle_t> _handle;
-      void* _result_row_address = nullptr;
+      void* _result_row_address{nullptr};
 
     public:
       bind_result_t() = default;
-      bind_result_t(const std::shared_ptr<detail::prepared_statement_handle_t>& handle) : _handle(handle)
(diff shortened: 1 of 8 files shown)


include/sqlpp11/mysql/bind_result.h
```diff
@@ -44,11 +44,11 @@ namespace sqlpp
     class bind_result_t
     {
       std::shared_ptr<detail::prepared_statement_handle_t> _handle;
-      void* _result_row_address = nullptr;
+      void* _result_row_address {nullptr};
```
rbock, line 47: Here in the other places, please omit the space between variable and brace initializer (clang-format).

MeanSquaredError: OK, I removed the spaces, run it through clang-format just to confirm that it no longer complains and then re-ran the tests again. It should be ready for a review again.

rbock: Thanks!

rbock:
LGTM, thanks!

PR #511, opened 2023-08-07
Cleanup the code of the MySQL connector

MeanSquaredError:
This PR makes the following changes to the PostgreSQL connector code:
- The following classes had their method definitions partly in-class and partly out-of-class:
 ```
sqlpp::postgresql::detail::statement_handle_t
sqlpp::postgresql::detail::prepared_statement_handle_t
sqlpp::postgresql::prepared_statement_t
sqlpp::postgresql::::Result
sqlpp::postgresql::bind_result_t
sqlpp::postgresql::connection_base
```
For the above classes all method definitions were moved in-class.

- Several methods that were defined in-class had superfluous `inline` specifiers that were removed.
- Removed `sqlpp::postgresql::Result::hasError()` which was declared but not defined and not used.
- Added/removed a couple of comments.
- Simplified cleanup of native PostgreSQL handles pretty much like we did with MySQL handles.
- Added `const` qualifiers.
- Removed superfluous namespace qualifiers.
- Changed some variables/methods from CamelCase to snake_case.

The PR has been built with
```
```

### DIFF
--- a/include/sqlpp11/postgresql/bind_result.h
+++ b/include/sqlpp11/postgresql/bind_result.h
@@ -64,16 +64,16 @@ namespace sqlpp
         }
 
         // Fetch total amount
-        if (_handle->totalCount == 0U)
+        if (_handle->total_count == 0U)
         {
-          _handle->totalCount = _handle->result.records_size();
-          if (_handle->totalCount == 0U)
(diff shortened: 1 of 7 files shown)


rbock:
Thanks for the large-scale cleanup!

PR #515, opened 2023-08-13
MySQL: const std::string -> const std::string&

MeanSquaredError:
Minor cleanup of the MySQL connector. 
Change a parameter type from `const std::string` to `const std::string&`


MeanSquaredError:
Github and/or Appveyor were having some kind of transient problem and the Appveyor tests failed. So I force pushed the same PR again, just to force Appveyor to run the tests again.

rbock:
Thanks!

PR #517, opened 2023-08-19
Unify date/time value parsing in connectors

MeanSquaredError:
This PR unifies the date/time-parsing code in the PostgreSQL, MySQL and SQLite3 connectors using the code in the PostgreSQL connector as basis for the new common functions.

Essentially I moved the PostgreSQL date/time-parsing code into several common functions in the `sqlpp::detail` namespace, placed them in `include/sqlpp11/detail/parse_date_time.h` and then replaced the custom date/time-parsing code in the MySQL and SQLite3 connectors with calls to these common functions. The SQLite3 connector still has one custom date/time parsing function because its date_time data fields are a bit special in that they can hold either date_time fields or date fields.

@rbock I am not sure if `include/sqlpp11/detail/parse_date_time.h` is the proper location for these common functions, so if you think that there is a better place for them, then I can adjust the PR.

This PR was built and tested with
```
cmake --build build -- VERBOSE=1
cd build
ctest
```

All tests passed successfully

### DIFF
--- a/include/sqlpp11/detail/parse_date_time.h
+++ b/include/sqlpp11/detail/parse_date_time.h
@@ -1,0 +1,128 @@
+#pragma once
+
+/**
+ * Copyright (c) 2023, Vesselin Atanasov
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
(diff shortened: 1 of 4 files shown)


rbock:
Thanks for unifying the date/time parsing. LGTM.

PR #518, opened 2023-08-22
A couple of minor cleanups in the PostgreSQL connector

MeanSquaredError:
This PR performs a couple of minor cleanups in the PostgreSQL connector, which I missed in the previous PRs:

- Replace ()-init with {}-init in the member initializer list of `sqlpp::postgresql::detail::connection_handle `
- Move the method definitions for `sqlpp::postgresql::detail::connection_handle` inside the class


rbock:
Thanks, LGTM!

PR #521, opened 2023-08-30
Move method definitions inside corresponding classes

MeanSquaredError:
This PR makes a few minor cleanups of the connection pool code. 

The method definitions of
```
sqlpp::connection_pool
sqlpp::connection_pool::pool_core
sqlpp::pooled_connection
sqlpp::normal_connection
```
are moved inside their corresponding classes. Also when I originally wrote the code in these methods, I used K&R indentation and the rest of the project seems to use Allman indentation so now I changed the indentation of the above classes to Allman-style.

Essentially I changed the code from
```
if (cond) {
    ...
} else {
    ...
}
```

to
```
if (cond)
{
    ...
}
else
{
    ...
}
```

As usual the PR was built and tested with 
```
cmake --build build
cd build
ctest
```
All tests passes successfully.

### DIFF
--- a/include/sqlpp11/connection.h
+++ b/include/sqlpp11/connection.h
@@ -46,8 +46,17 @@ namespace sqlpp
 
     // Constructors
     normal_connection() = default;
-    normal_connection(const _config_t& config);
-    normal_connection(const _config_ptr_t& config);
+
+    normal_connection(const _config_t& config) :
+      normal_connection{std::make_shared<_config_t>(config)}
+    {
+    }
+
+    normal_connection(const _config_ptr_t& config) :
+      ConnectionBase{std::make_unique<_handle_t>(config)}
+    {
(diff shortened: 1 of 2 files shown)


rbock:
Thanks for the cleanup! Did you use clang-format to make the format changes?

MeanSquaredError:
@rbock,
I just changed the indentation manually because there were only 4 or 5 places in the connection pool code where the braces were using K&R style.

Do you want me to run the two files through clang-format and push the changes?

rbock:
Yes, as you are looking into formatting anyways, it would make sense to run the whole file, I think.

MeanSquaredError:
OK, I ran `connecton.h` and `connection_pool.h` through clang-format.

rbock:
Thanks for the cleanup!

Issue #523, opened 2023-09-12
Compilation error with c++11 and gcc 4.8.5


Piogar:
Hello,

I tried to compile library with c++11 (centos, gcc 4.8.5, clean repo without any modifications, just cmake and make run) and faced following error:
`[  2%] Building CXX object tests/core/usage/CMakeFiles/sqlpp11_core_tests.dir/BooleanExpression.cpp.o
In file included from /root/tmp/sqlpp/sqlpp11/tests/core/usage/MockDb.h:29:0,
                 from /root/tmp/sqlpp/sqlpp11/tests/core/usage/BooleanExpression.cpp:28:
/root/tmp/sqlpp/sqlpp11/include/sqlpp11/connection.h: In constructor 'sqlpp::pooled_connection<ConnectionBase>::pooled_connection(const _config_ptr_t&, sqlpp::pooled_connection<ConnectionBase>::_pool_core_ptr_t)':
/root/tmp/sqlpp/sqlpp11/include/sqlpp11/connection.h:124:26: error: 'make_unique' is not a member of 'std'
         : ConnectionBase{std::make_unique<_handle_t>(config)}, _pool_core{pool_core}`

Error seems obvious as std::make_unique is available in c++14 whereas gcc I have to use supports only c++11.

Is that expected and I'm missing something or is it a bug making building with c++11 impossible?

MeanSquaredError:
It is a bug introduced by one of my PRs. Can you please try this branch:
https://github.com/MeanSquaredError/sqlpp11/tree/fix_make_unique
It should build with C++11. If it works correctly for you then I will open a PR with the fix.

Piogar:
Thanks for quick answer. Commit seems to resolve this issue, however now I face issues #431 . I'll check #431 thread for details.

MeanSquaredError:
@Piogar 
Opening an issue here (what you already did) is probably the fastest way to get support.
There is also a chat on gitter:
https://gitter.im/sqlpp11/Lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
but it seems to be less active than the support here.

MeanSquaredError:
@Piogar 
Regarding issue #431
It doesn't seem to be one of the things that I broke with my patches :-)

[...]

PR #524, opened 2023-09-13
Replace std::make_unique<T>(..) with std::unique_ptr{new T{...}}

MeanSquaredError:
This PR replaces std::make_unique<T>(..) with std::unique_ptr{new T{...}}
This change fixes https://github.com/rbock/sqlpp11/issues/523

### DIFF
--- a/include/sqlpp11/connection.h
+++ b/include/sqlpp11/connection.h
@@ -51,7 +51,7 @@ namespace sqlpp
     {
     }
 
-    normal_connection(const _config_ptr_t& config) : ConnectionBase{std::make_unique<_handle_t>(config)}
+    normal_connection(const _config_ptr_t& config) : ConnectionBase{std::unique_ptr<_handle_t>{new _handle_t{config}}}
     {
     }
 
@@ -65,7 +65,7 @@ namespace sqlpp
     // creates a connection handle and connects to database
     void connectUsing(const _config_ptr_t& config) noexcept(false)
     {
-      ConnectionBase::_handle = std::make_unique<_handle_t>(config);
+      ConnectionBase::_handle = std::unique_ptr<_handle_t>{new _handle_t{config}};
     }
 
   private:
@@ -121,7 +121,7 @@ namespace sqlpp
     }
 
     pooled_connection(const _config_ptr_t& config, _pool_core_ptr_t pool_core)
-        : ConnectionBase{std::make_unique<_handle_t>(config)}, _pool_core{pool_core}
+        : ConnectionBase{std::unique_ptr<_handle_t>{new _handle_t{config}}}, _pool_core{pool_core}
     {
     }
 
--- a/tests/include/ConnectionPoolTests.h
+++ b/tests/include/ConnectionPoolTests.h
@@ -236,7 +236,7 @@ namespace sqlpp
       void test_destruction_order(typename Pool::_config_ptr_t config)
       {
         // Create a pool, get a connection from it and then destroy the pool before the connection
-        auto pool = std::make_unique<Pool>(config, 5);
+        auto pool = std::unique_ptr<Pool>{new Pool{config, 5}};
         auto conn = pool->get();
         pool = nullptr;
       }


MeanSquaredError:
@rbock 
This PR is a quick fix.

However IMHO it would be nice to use a `make_unique` when it is supported because:
1.  `make_unique<T>()` is simpler to write and maintain vs. `unique_ptr<T>{new T}`
2.  `make_unique<T>()`  does one less memory allocation vs `unique_ptr<T>{new T}`.

So in the long run maybe it is a good idea to add a namespace `sqlpp::compat` and put a custom implementation of `make_unique` there, i.e.
`sqlpp::compat::make_unique` and then use it throughout the code?

Also if in the future a custom version of `std::optional` is implemented, we could also add `sqlpp::compat::optional` there.

What do you think?

CJCombrink:
My 2c: I am in favour of a `sqlpp::compat` with a custom/manual implementation for the few things mentioned if the compiler does not have it, and then an alias to the proper thing if the compiler has it. `std::optional` and `std::make_unique` have feature test macros for this type of solution.

That way it will be a drop in/out replacement for code bases when they migrate.

CJCombrink:
> 2\. `make_unique<T>()`  does one less memory allocation vs `unique_ptr<T>{new T}`.

3. `make_unique<T>()` have less chance to leak if the constructor arguments throw in `unique_ptr<T>{new T(this_throws())};`

MeanSquaredError:
> My 2c: I am in favour of a `sqlpp::compat` with a custom/manual implementation for the few things mentioned if the compiler does not have it, and then an alias to the proper thing if the compiler has it.

Also, thinking about it, `sqlpp::compat` will make things simpler when the times comes to port the code to sqlpp26. Then it will be possible to just drop `sqlpp::compat` and rename `sqlpp::compat::xxxxx` to `std::xxxxx` in the source code.

> `std::optional` and `std::make_unique` have feature test macros for this type of solution.

I just looked into the feature test macros and it seems that they are more or less reliable only if the compiler supports `C++20`. The compiler can be used in C++11,C++14 or C++17 mode but only if it supports C++20, the feature test macros will be reliable.

Older versions of GCC and MSVC that are C++11-compiant don't seem to support the feature test macros altogether, or at least some of the macros are missing.
https://stackoverflow.com/questions/49316752/feature-test-macros-not-working-properly-in-visual-c
https://stackoverflow.com/a/76478005

So I am thinking about using something like this:
```
#ifdef _MSVC_LANG
#define CXX_STD_VERSION _MSVC_LANG
#else
#define CXX_STD_VERSION __cplusplus
#endif

namespace sqlpp
{
    namespace compat
    {

#if CXX_STD_VERSION >= 201402L
        using make_unique = std::make_unique;
#else
        template<typename T, typename... Args>
        std::unique_ptr<T> make_unique(Args&&... args)
        {
            return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
        }
#endif
    }  // namespace compat
}  // namespace sqlpp
```

It may need to be a bit more complex depending on whether we want to support VS2012 
https://stackoverflow.com/questions/12547983/is-there-a-way-to-write-make-unique-in-vs2012

But in general, I think that it should be something along the lines of the above implementation.

Any comments and/or suggestions are welcome!

CJCombrink:
It looks good.

I also support the `_MSVC_LANG` define for MSVC since we recently fell in this trap when working with `std::string_view`. We thought it was a bug in sqlpp11, until we found the following post: [MSVC now correctly reports __cplusplus](https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/). We build on gcc and MSVC and some code compiled only on GCC...

rbock:
+1

Sounds reasonable to me.

MeanSquaredError:
OK, then I will prepare shortly a PR with the proposed `sqlpp::compat::make_unique`

MeanSquaredError:
As per our discussion I have added a new PR that adds support for `sqlpp::compat::make_unique`. I am closing this PR as it is obsolete.

PR #527, opened 2023-09-14
Add sqlpp::compat::make_unique

MeanSquaredError:
As per our discussion here https://github.com/rbock/sqlpp11/pull/524 , this PR adds support for `sqlpp::compat::make_unique` and replaces all calls to `std::make_unique` with `sqlpp::compat::make_unique`.

The PR was build and tested with g++ in C++11 mode with
```
cmake -B build -DBUILD_POSTGRESQL_CONNECTOR=ON -DBUILD_SQLITE3_CONNECTOR=ON -DBUILD_MYSQL_CONNECTOR=ON -DBUILD_TESTING=ON -DUSE_SYSTEM_DATE=ON -DDEPENDENCY_CHECK=ON -DSQLPP11_TESTS_CXX_STD=11
```

And then it was built and tested with g++ in C++17 mode with
```
```


MeanSquaredError:
By the way, I did not add support for VS 2012 in `sqlpp::compat::make_unique`, because from what I saw online, VS 2012 does not support template argument packs (`typename... Args`) and sqlpp11 uses them in quite a few places, so most likely sqlpp11 would not work with VS 2012 anyway.

include/sqlpp11/compat/make_unique.h
```diff
@@ -0,0 +1,51 @@
...
+#else
+#define CXX_STD_VER __cplusplus
+#endif
+
+namespace sqlpp
+{
+  namespace compat
+  {
+    template <typename T, typename... Args>
+    inline std::unique_ptr<T> make_unique(Args&&... args)
```
rbock, line 42: You should be able to use a function alias for C++14 and above, see for instance:

https://www.fluentcpp.com/2017/10/27/function-aliases-cpp/

MeanSquaredError: I also thought about using a templated variable, but unfortunately function aliases implemented via templated variables, don't support type deduction, so for `make_unique` we would have to specify the types for all the parameters which makes it more difficult to use compared to the original `std::make_unique`. Below is a sample program which shows the problem
```
#include <memory>

template <typename T, typename... Args>
const auto custom_make_unique = std::make_unique<T, Args...>;

int main()
{
	// Works fine with C++14 and newer
	[[maybe_unused]] auto var_1 = std::make_unique<std::pair<int, int>> (1, 2);

	// Works fine with C++14 and newer
	[[maybe_unused]] auto var_2 = custom_make_unique<std::pair<int, int>, int, int> (1, 2);

	// Fails with C++14 and newer, with "error: too many arguments to function"
	// because templated function aliases via templated variabled don't support type deduction
	[[maybe_unused]] auto var_3 = custom_make_unique<std::pair<int, int>> (1, 2);
	return 0;
}

```

MeanSquaredError:
Added an extra test which calls `sqlpp::compat::make_unique` with 2 arguments.
Rebased against the latest main branch.

include/sqlpp11/compat/make_unique.h
rbock: That's sad, but we'll have to live with it, I guess. Thanks for looking into it!

rbock, line 42: Please remove the `inline` (not needed for a template function).

MeanSquaredError: The purpose of the `inline` specifier  here was not to avoid the duplicate function definitions but rather to hint the compiler to inline the instantiated function and thus avoid the extra function call. There was no guarantee that it would work but I thought that maybe it would increase the chance of inlining.

But I removed it anyway.

rbock: Thanks, and see http://www.gotw.ca/gotw/033.htm for instance.

rbock:
Thanks for fixing this!

PR #535, opened 2023-10-12
Fix warning for `greater_equal` operator.

korolew:
Call `verify()` for `greater_equal` operator to fix `warning: expression result unused [-Wunused-value]`.

### DIFF
--- a/include/sqlpp11/basic_expression_operators.h
+++ b/include/sqlpp11/basic_expression_operators.h
@@ -171,7 +171,7 @@ namespace sqlpp
     auto operator>=(T t) const -> _new_binary_expression_t<greater_equal_t, T>
     {
       using rhs = wrap_operand_t<T>;
-      check_comparison_t<Expr, rhs>{};
+      check_comparison_t<Expr, rhs>::verify();
 
       return {*static_cast<const Expr*>(this), rhs{t}};
     }


rbock:
Thanks for the fix. Which compiler are you using?

 I just tried to reproduce the warning and failed with clang-14 and gcc-12 and  `-Wall -Wextra -Wpedantic`

e.g. I added this block to tests/core/serialize/Operator.cpp

```
  // Comparison
  compare(__LINE__, bar.alpha < 3u, "(tab_bar.alpha<3)");
  compare(__LINE__, bar.alpha <= 3u, "(tab_bar.alpha<=3)");
  compare(__LINE__, bar.alpha == 3u, "(tab_bar.alpha=3)");
  compare(__LINE__, bar.alpha != 3u, "(tab_bar.alpha<>3)");
  compare(__LINE__, bar.alpha >= 3u, "(tab_bar.alpha>=3)");
  compare(__LINE__, bar.alpha > 3u, "(tab_bar.alpha>3)");
```

MeanSquaredError:
> I just tried to reproduce the warning and failed with clang-14 and gcc-12 and `-Wall -Wextra -Wpedantic`

Just for the record, I haven't seen this warning either with sqlpp11 and GCC 13.2.1. It seems that the big three (GCC, Clang, MSVC) normally don't complain about created but otherwise unused instances of classes/structures because they have constructors/destructors which are invoked upon construction/destruction of the object.

I have seen similar warnings being triggered by arcane combinations of the `-g` and/or `-O` options.

korolew:
I'm using clang-17 with CMAKE_CXX_STANDARD 20 enabled in my project.

rbock:
Interesting, clang-14 also yields the warning in C++20 mode.

I wonder why that's standard-specific?

Anyways, thanks for the fix! Merging.

PR #545, opened 2023-11-13
Fix warnings in ddl2cpp

MeanSquaredError:
There are a three regular expressions in `ddl2cpp` which lack proper escaping and  Python 3.12.0 generates warnings.
```
/usr/local/bin/sqlpp11-ddl2cpp:475: SyntaxWarning: invalid escape sequence '\s'
  return re.sub("(^|\s|[_0-9])(\S)", repl_camel_case_func, s)
/usr/local/bin/sqlpp11-ddl2cpp:480: SyntaxWarning: invalid escape sequence '\s'
  return re.sub("(\s|_|[0-9])(\S)", repl_camel_case_func, s)
/usr/local/bin/sqlpp11-ddl2cpp:494: SyntaxWarning: invalid escape sequence '\s'
  var_name = first_lower(re.sub("(\s|-|[0-9])(\S)", repl_func_for_args, s))
```

 Right now it is a minor issue as Python automatically adds the missing backslashes to the regular expressions, but recently they changed these warnings from `DeprecationWarning` to `SyntaxWarning` and it seems that they are planning to make them errors further down the road.

This PR fixes the warnings by adding the missing backslashes.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -472,12 +472,12 @@ def repl_camel_case_func(m):
 
 def class_name_naming_func(s):
     s = s.replace(".", "_")
-    return re.sub("(^|\s|[_0-9])(\S)", repl_camel_case_func, s)
+    return re.sub(r"(^|\s|[_0-9])(\S)", repl_camel_case_func, s)
 
 
 def member_name_naming_func(s):
     s = s.replace(".", "_")
-    return re.sub("(\s|_|[0-9])(\S)", repl_camel_case_func, s)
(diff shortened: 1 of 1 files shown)


rbock:
Thanks!
Alternatively, we could turn the strings into raw strings, I guess? Might be better for readability.
What do you think?

MeanSquaredError:
@rbock 
I agree that in this case a raw string is simpler, so I updated the PR to use raw strings.

rbock:
Nice, thanks!

PR #550, opened 2024-01-03
Add is_transaction_active() to all connectors

MeanSquaredError:
Currently the MySQL connector provides a handy method called `is_transaction_active()` which allows the user to check if the database connection is running a transaction, without making a round trip to the server and without using database-specific commands or API calls. The PostgreSQL and SQLite3 connectors keep internally information about the transaction state, but do not allow the user to query the transaction state.

This PR unifies the connector API by adding a public method `is_transaction_active()` to all the connectors.

The internal state of the SQLite3 connector supports 3 values, `tx active`, `tx not active` and `maybe`, where `maybe` is the state set while the transaction is being started/stopped. However these connector methods are not re-entrant or multithread-safe and the SQLite3 state variable is not atomic, so it does not make sense to keep a transient "maybe" state which is used only inside the transaction methods without being exported outside to the other methods or to the user. That is why this PR replaces the SQLite3 transaction state with a boolean variable, just like the one used for MySQL and PostgreSQL.

The PR also adds tests for `is_transaction_active()`

The PR was built and tested with
```
cmake --build build
cd build
ctest
```

### DIFF
--- a/connector_api/connection.h
+++ b/connector_api/connection.h
@@ -203,6 +203,9 @@ namespace sqlpp
       //! report a rollback failure (will be called by transactions in case of a rollback failure in the destructor)
       void report_rollback_failure(const std::string message) noexcept;
 
+      //! check if transaction is active
+      bool is_transaction_active();
+
     protected:
       // Low-level connection handle
       _handle_ptr_t _handle;
--- a/connector_api/connection.h
+++ b/connector_api/connection.h
@@ -180,7 +180,6 @@ namespace sqlpp
         return t._prepare(*this);
       }
 
-
       //! set the transaction isolation level for the current connection
       /// time of effect is connector-specific, for most is will only affect new transactions
       void set_default_isolation_level(sqlpp::isolation_level);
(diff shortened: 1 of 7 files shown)


rbock (changes requested):
Looks like you are modeling the strategy for setting `_transaction_active = false` after the mysql connector.
I wonder if that is correct? The transaction is certainly still active before sending `COMMIT`?

MeanSquaredError:
OK, I changed the transaction shutdown sequence to:
1. Report transaction shutdown. This is done for rollback only.
2. Execute SQL statement that shuts down the transaction.
3. Sets the _transaction_active flag to false.

Originally I used the MySQL order in all connectors just to have consistent order in all connectors. My rationale for using the MySQL order was that it didn't really matter since the connector methods were not re-entrant, so what really mattered was the value of `_transaction_active` before and after the method call.

But after your comment I though that you are right, the MySQL order is wrong, because an exception could occur while executing the SQL statement that stops the transaction and then connection object would be left in invalid state, the transaction would be active, but `_transaction_active` would be false. So I changed everything to execute the SQL statement before clearing the transaction flag.

rbock (approved):

rbock:
Thanks for the update! Looks good to me.

PR #551, opened 2024-01-04
PostgreSQL exceptions for serialization_failure and deadlock_detected

MeanSquaredError:
This PR adds exception classes for PostgreSQL errors

40001 (serialization_failure)
40P01 (deadlock_detected)

Handling these error codes is useful when implementing optimistic locking where transactions are expected to fail sometimes with certain error codes and such transactions should be retried (semi-)automatically.

Overall the exception hierarchy of sqlpp needs an overhaul to allow a common implementation that would cover all connectors at once. This PR is more or less a temporary band-aid to make optimistic locking easier to implement with PostgreSQL.

### DIFF
--- a/include/sqlpp11/postgresql/exception.h
+++ b/include/sqlpp11/postgresql/exception.h
@@ -299,6 +299,17 @@ namespace sqlpp
       }
     };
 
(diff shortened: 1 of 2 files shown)


rbock:
Thanks, looks reasonable to me.

PR #552, opened 2024-01-05
Missing break when throwing PostgreSQL transaction exceptions

MeanSquaredError:
The previous PR that added support for PostgreSQL transaction exceptions (https://github.com/rbock/sqlpp11/pull/551) lacks a break statement in the newly added switch case. This does not cause any bugs but leads to slightly inefficient code and causes compiler warnings about the unintentional fall-through. 

I am not sure why I missed the warning previously, but here is the PR that fixes it. As usual it was built and tested with
```
cmake --build build
cd build
ctest
```


rbock:
Ooops, thanks!

PR #567, opened 2024-04-09
Update FindMariaDB.cmake

JoyBinY:
Users can use 
-DCMAKE_LIBRARY_PATH=${INSTALL}/lib
 without needing to specify mariadb separately like 
-DCMAKE_LIBRARY_PATH=${INSTALL}/lib/mariadb


rbock:
Thanks!

PR #572, opened 2024-06-11
Recognize serial2, serial4 and serial8 types in ddl2cpp

MeanSquaredError:
While trying to make primary key columns NOT NULL, I came across another limitation of ddl2cpp. It does not recognize the PostgreSQL-specific data types serial2, serial4 and serial8.

When it comes across an unknown column type (like one of these serial types), ddl2cpp ignores any column modifiers like NOT NULL, PRIMARY KEY, etc.

This is a quick fix that adds support for the three data types.

One might argue that the correct behavior when coming across an unknown data type would be just to terminate the script with error, but this could break things  in the future when some of the databases adds a new data type or an alias for an existing data type.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -92,6 +92,9 @@ ddlIntegerTypes = [
 ddlSerialTypes = [
     "bigserial",  # PostgreSQL
     "serial",  # PostgreSQL
+    "serial2",  # PostgreSQL
+    "serial4",  # PostgreSQL
+    "serial8",  # PostgreSQL
     "smallserial",  # PostgreSQL
 ]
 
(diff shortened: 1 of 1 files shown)


rbock:
Thanks for adding these data types!

And yes, terminating would be an option, or at least printing a warning (with the suggestion to report a bug about it).

MeanSquaredError:
I investigated the issue in more detail. It turned out that ddl2cpp already has some code to print and error and terminate if unknown data type is found. 

However there are a bunch of places where incorrectly `pp.CaselessLiteral` is used instead of the correct `pp.CaselessLiteral`, so for example if the user specifies the non-existent type `serial5`, the real type `serial` matches and `5` is left as a remainder which messes up the parsing, so ddl2cpp does not err, but produces bad header data.

I am currently preparing a fix for those issues.

PR #574, opened 2024-06-11
Fix SQL keyword matching in ddl2cppFix ddl2cpp keywords

MeanSquaredError:
I deleted by mistake the branch of the previous PR, so I am re-submitting it.

Normally ddl2cpp terminates with an error message if the input .sql script has a column with an unknown data type. However it does not err if there is a valid data type which is a prefix of the unknown data type.

For example ddl2cpp will err for the unknown data type `invalidtype`, but will not err for `serial5`, because its prefix, `serial`, is a valid data type. In the latter case the ddl2cpp parse will consume the prefix `serial` while the remainder `5` will not cause an error, but will cause the parser to ignore valid modifiers, like `NOT NULL`, `DEFAULT`, etc.

The cause of the bug is that ddl2cpp uses `pp.CaselessLiteral` to match keywords, while it should actually use `pp.CaselessKeyword`. The former matches any valid prefix, not paying attention to word boundaries. While the latter matches only whole words.

This PR fixes the bug and causes dddl2cpp to always err for invalid data types and it also fixes other similar problems caused by the use of `pp.CaselessLiteral`. The PR also improves the testing for invalid data types.

### DIFF
--- a/scripts/ddl2cpp
+++ b/scripts/ddl2cpp
@@ -45,7 +45,8 @@ ddlString = (
 ddlTerm = pp.Word(pp.alphas + "_", pp.alphanums + "_.$")
 ddlName = pp.Or([ddlTerm, ddlString, pp.Combine(ddlString + "." + ddlString), pp.Combine(ddlTerm + ddlString)])
 ddlOperator = pp.Or(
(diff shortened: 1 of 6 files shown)


rbock:
Thanks a lot for the investigation and the PR. Looks good to me.

PR #575, opened 2024-06-12
ddl2cpp: primary key column cannot be null

MeanSquaredError:
This PR updates ddl2cpp so that columns with the `PRIMARY KEY` constraint never get the `sqlpp::tag::can_be_null` tag.

The PR only affects `PRIMARY KEY` constraints on column level. A `PRIMARY KEY` constraint defined on a table level (after column definitions), still can get the `sqlpp::tag::can_be_null` tag. In order to fix that for table-level constraints more changes to the codebase would be needed.


rbock:
Nice, thanks for the fix!

Issue #591, opened 2024-08-31
Suspicious implementation of result_t::iterator::operator++(int)


MeanSquaredError:
After delving into the code that implements result iterators I found the implementation of `result_t::iterator::operator++(int)` a bit suspicious.

In include/sqlpp11/result.h on line 122 we have the following code
```
      iterator& operator++()
      {
        _result.get().next(_result_row.get());
        return *this;
      }

      iterator operator++(int)
      {
        auto previous_it = *this;
        _result.next(_result_row.get());
        return previous_it;
      }
```
Obviously the purpose of these two methods is to implement the prefix increment (`++it`) and postfix incremenet (`it++`) operators. However the problem is with this line:
```
        auto previous_it = *this;
```
Both `previous_it` and `*this` actually reference the same underlying result object, so the two iterators essentially point to the same row in the results sequence.

I don't see an easy fix for this without making intrusive changes to the code, but it is probably worth adding a comment to let users know that the postfix increment operator is actually a prefix increment operator.

rbock:
Good find. Adding documentation seems to be the best option.

And maybe also implementing `iterator operator++(int)` by calling `iterator& operator++()`.

MeanSquaredError:
I provided a PR, but made it into a draft, because I am not sure if we actually need the post-increment. Please see my comment in the PR.

MeanSquaredError:
I guess this issue is resolved, at least for the time being, so I will close it.

PR #592, opened 2024-08-31
Comment a limitation of result_t::iterator::operator++(int)

MeanSquaredError:
As per our discussion here https://github.com/rbock/sqlpp11/issues/591 this PR makes result_t::iterator::operator++(int) an alias of result_t::iterator::operator++() and add a comment about its limitation.

The PR was built and tested with:
```
cmake --build build
cd build
ctest
```

### DIFF
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -127,9 +127,8 @@ namespace sqlpp
 
       iterator operator++(int)
       {
-        auto previous_it = *this;
-        _result.next(_result_row.get());
-        return previous_it;
+        // FIXME: Currently the postfix increment operator does a prefix increment
+        return ++*this;
       }
 
       std::reference_wrapper<db_result_t> _result;


MeanSquaredError:
Actually thinking again about this problem, maybe it is better just to remove the postfix increment operator? According to the description here https://en.cppreference.com/w/cpp/language/range-for the range-based for loop seems to be using a preincrement, so the postincrement is unnecessary and just broken.

What do you think?

MeanSquaredError:
I just looked into the requirements of the `std::input_or_output_iterator` concept which is required for a class to be a range/view. It seems that it requires a postfix increment operator, but the operator does not have to return any value, that is it can be `void`.

What do you think about changing the postfix increment operator to return `void`, instead of fixing it (which is pretty difficult given the nature of the underlying database result) or removing it.

It seems to be a good compromise that will allow use of the results as ranges. What do you think?

MeanSquaredError:
I am closing this PR, because it has been replaced by https://github.com/rbock/sqlpp11/pull/593

PR #593, opened 2024-08-31
Turn result_t into a C++20 range

MeanSquaredError:
This PR turns `result_t` objects into C++20 ranges that can be used with various range/view algorithms. Since C++20 views are essentially movable ranges and `result_t` is movable, this means that `result_t` is also a view.

C++20 input ranges are single-pass, i.e. they are only required to support a single call to `begin()` and range algorithms are prohibited from calling `begin()` a second time on an input range. This requirement fits quite well the way sqlpp11's database results behave, because they are single-pass too.

The PR makes two main changes to `result_t::iterator`

- The return type of the postfix increment operator is changed to `void`. To facilitate the use of postfix increment with input ranges, their postfix increment operator is allowed (and actually recommended) to return `void`.
- The iterator is made default-constructible. To achieve that its references to the result data and result row are changed to pointers. For default-constructed iterators these pointers are set to `nullptr`.

The PR also adds a comment about `result_t::iterator::operator==`, which allows comparison of iterators that don't belong to the same sequence (i.e. database result). Such comparison is undefined behavior by the standard, but the old code allowed it. The updated code also allows that, but I added a comment about it. It should probably be fixed further down the road, but such a fix would be too intrusive, so I decided just to document it.

### DIFF
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -81,7 +81,14 @@ namespace sqlpp
     class iterator
     {
     public:
+#if __cplusplus >= 202002L
+      using iterator_concept = std::input_iterator_tag;
+#else
+      // LegacyInputIterator describes best our iterator's capabilities. However our iterator does not
+      // really fulfil the requirements for LegacyInputIterator because its post-increment operator
+      // returns void.
       using iterator_category = std::input_iterator_tag;
+#endif
       using value_type = result_row_t;
       using pointer = const result_row_t*;
       using reference = const result_row_t&;
@@ -133,10 +140,8 @@ namespace sqlpp
 
       // It is quite difficult to implement a postfix increment operator that returns the old iterator
       // because the underlying database results work in a stream fashion not allowing to return to
-      // previously-read rows.
-      //
-      // We need the postfix increment mostly for compatibility with C++20 ranges so we set the return
-      // type to "void" which is allowed for C++20 range iterators.
+      // previously-read rows. That is why we set the post-increment return type to void, which is
+      // allowed for C++20 input iterators
       //
       void operator++(int)
       {
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -93,6 +93,11 @@ namespace sqlpp
       using reference = const result_row_t&;
       using difference_type = std::ptrdiff_t;
 
+      iterator()
+          : _result_ptr(nullptr), _result_row_ptr(nullptr)
+      {
+      }
+
       iterator(db_result_t& result, result_row_t& result_row)
           : _result_ptr(&result), _result_row_ptr(&result_row)
       {
@@ -110,6 +115,14 @@ namespace sqlpp
 
       bool operator==(const iterator& rhs) const
       {
+        if ((_result_row_ptr != nullptr) != (rhs._result_row_ptr != nullptr))
+        {
+          return false;
+        }
+        if (_result_row_ptr == nullptr)
+        {
+          return true;
+        }
         return *_result_row_ptr == *rhs._result_row_ptr;
       }
 
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -32,12 +32,6 @@
 
 namespace sqlpp
 {
-  template <typename>
-  struct iterator_category
-  {
-    using type = std::input_iterator_tag;
-  };
-
   namespace detail
   {
     template<class DbResult, class = void>
@@ -87,7 +81,7 @@ namespace sqlpp
     class iterator
     {
     public:
-      using iterator_category = typename iterator_category<DbResult>::type;
+      using iterator_category = std::input_iterator_tag;
       using value_type = result_row_t;
       using pointer = const result_row_t*;
       using reference = const result_row_t&;
--- a/include/sqlpp11/result.h
+++ b/include/sqlpp11/result.h
@@ -93,25 +93,24 @@ namespace sqlpp
       using reference = const result_row_t&;
       using difference_type = std::ptrdiff_t;
 
-      iterator(std::reference_wrapper<db_result_t> result,
-               std::reference_wrapper<result_row_t> result_row) :
-          _result(std::move(result)), _result_row(std::move(result_row))
+      iterator(db_result_t& result, result_row_t& result_row)
+          : _result_ptr(&result), _result_row_ptr(&result_row)
       {
       }
 
       reference operator*() const
       {
-        return _result_row;
+        return *_result_row_ptr;
       }
 
       pointer operator->() const
       {
-        return &_result_row.get();
+        return _result_row_ptr;
       }
 
       bool operator==(const iterator& rhs) const
       {
-        return _result_row.get() == rhs._result_row.get();
+        return *_result_row_ptr == *rhs._result_row_ptr;
       }
 
       bool operator!=(const iterator& rhs) const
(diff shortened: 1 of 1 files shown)


include/sqlpp11/result.h
```diff
@@ -93,25 +93,39 @@ namespace sqlpp
...
+        if ((_result_row_ptr != nullptr) != (rhs._result_row_ptr != nullptr))
+        {
+          return false;
+        }
+        if (! _result_row_ptr)
+        {
+          return true;
+        }
+        // FIXME: Throw an exception when someone tries to compare iterators from different sequences,
+        // because this is undefined behavior.
```
rbock, line 127: "undefined behavior" means undefined behavior, not "throw an exception" or "crash" or "abort" or anything.

Returning a result like this is perfectly fine, afaict.

rbock (commented):
Thanks for the PR!
Looks good to me except for the `FIXME` which I think is superfluous.

include/sqlpp11/result.h
MeanSquaredError: OK, I removed the comment.

Indeed it is perfectly fine to do anything on undefined behavior including the current action. However I think that comparing iterators from different sequences is clearly a programming mistake and it is difficult to imagine a scenario where it would be done intentionally. That is why I added the comment.

include/sqlpp11/result.h
```diff
@@ -93,25 +93,37 @@ namespace sqlpp
...
       }
 
       bool operator==(const iterator& rhs) const
       {
-        return _result_row.get() == rhs._result_row.get();
+        if ((_result_row_ptr != nullptr) != (rhs._result_row_ptr != nullptr))
+        {
+          return false;
+        }
+        if (! _result_row_ptr)
```
rbock, line 122: Sorry, one more nit: You are using two different ways of of testing for `nullptr`.

Can we use `== nullptr` throughout this function?

include/sqlpp11/result.h
```diff
@@ -121,29 +133,35 @@ namespace sqlpp
 
       iterator& operator++()
       {
-        _result.get().next(_result_row.get());
+        _result_ptr->next(*_result_row_ptr);
```
rbock, line 136: Note: This is undefined behavior for default-constructed iterators.

rbock (changes requested):

include/sqlpp11/result.h
MeanSquaredError: Assuming that `result_t::iterator` is modelling [LegacyInputIterator](https://en.cppreference.com/w/cpp/named_req/InputIterator), then that page says that for the expression `++r`, there is a precondition that the iterator `r` must be dereferenceable. Obviously default-constructed iterators are not dereferenceable, so calling `operator++` for a default-constructed iterator is undefined behavior per se. With the current code the most likely result will be segfault when the program tries to derefence the NULL pointer.

I think that it makes sense to leave it this way because it signals a bug in the caller's code. Alternatively we can either throw an exception, or just ignore the increment, the last approach IMHO being the worst one. What is your suggestion?

include/sqlpp11/result.h
MeanSquaredError: OK, I changed
```
if (! _result_row_ptr)
```
to
```
if (_result_row_ptr == nullptr)
```

include/sqlpp11/result.h
rbock: I am fine leaving it this way, just wanted to make sure this is not accidental.

include/sqlpp11/result.h
rbock: Thanks!

rbock:
Thanks for the conversation and changes. Happy to merge if you agree.

MeanSquaredError:
There are a couple of minor changes that I was thinking about:
In result.h on line 35 we have the following definition:
```
  template <typename>
  struct iterator_category
  {
    using type = std::input_iterator_tag;
  };
```
Then in the same file on line 90 we have
```
      using iterator_category = typename iterator_category<DbResult>::type;
```
I think it makes sense to replace that line with
```
      using iterator_category = std::input_iterator_tag;
```
and then get rid of the `struct iterator_category` at the top of the file.
What do you think?

MeanSquaredError:
Additionally I was reading the C++ standard on iterators and different discussions of the working group that standardized the iterators. It seems that there are two standards for iterators, the old (C++98/C++17 iterators) and the new ones (C++20).

The old iterators define the typedef `iterator_category` which has one of the following types:
```
input_iterator_tag
output_iterator_tag
forward_iterator_tag
bidirectional_iterator_tag
random_access_iterator_tag
```
The new C++20 iterators define the type `iterator_concept` which can have one of the above values plus one additional value of
```
contiguous_iterator_tag
```

The type names are the same for the old and the new iterators but the actual requirements (concepts) for the iterator types differ slightly. The most significant difference is in the requirements for the input iterators. The old requirements say that the post-increment must return the old value of the iterator. The new requirements allow the post-increment to have any return type including `void`. The change in the requirements is done because the old (C++98/C++17) requirement for is viewed as flowed because it is very difficult for one-pass iterators like the input iterator to implement the post-increment so that it would return the old iterator. So that is why they relaxed the requirement.

For such input iterators that are valid according to the C++20 requirements but invalid according to the C++17 requirements, the `input_category` should be left empty. If the C++20 iterator is also a valid C++17 iterator, then the `input_category` should be populated accordingly.

So the bottom line is that for the `result_t::iterator` we should add 
``
`using iterator_concept = std::input_iterator_tag
```
and remove the typedef for `iterator_category` because the iterator returns `void` for the post-increment operator and thus is not a valid C++17 iterator.

What do you think?

rbock:
> There are a couple of minor changes that I was thinking about: In result.h on line 35 we have the following definition:
> 
> ```
>   template <typename>
>   struct iterator_category
>   {
>     using type = std::input_iterator_tag;
>   };
> ```
> 
> Then in the same file on line 90 we have
> 
> ```
>       using iterator_category = typename iterator_category<DbResult>::type;
> ```

The rationale was: This does allow changing the iterator category based on `DbResult`. I don't think it has ever been used, though. So I am OK with removing.


> So the bottom line is that for the result_t::iterator we should add
> ```
> using iterator_concept = std::input_iterator_tag
> ```
> and remove the typedef for iterator_category because the iterator returns void for the post-increment operator and thus is not a valid C++17 iterator.

As long as this does not break algorithms in C++11 that assume input iterators, I am fine with that change. Alternativey, we could declare category/concept depending on the C++ standard.

MeanSquaredError:
OK, great I will update the PR by removing `struct iterator_category` and will add an `#ifdef` to add `iterator_concept` for C++20 or newer 
`iterator_category` for C++17 or older
the iterator_category is no longer really `std::input_iterator_tag`, but this is the closest that we can get for C++17.

My expectation is that this will not cause problems, because until now nobody really used `result_t` with ranges, and probably only used it with range-based `for` loops which don't care about categories or concepts as they only need `begin()` and `end()`.

MeanSquaredError:
OK, I made the changes, so I think the PR should be ready for inclusion unless you notice any other issues with the new code.

The PR uses the `__cplusplus` macro to detect the selected C++ standard version. However this macro does not work well on MSVC so on that compiler the iterator still defines `iterator_category` even when it compiles in C++20 or newer mode. It is not a big deal because on MSVC it behaves like it did before.

To detect the compiler version correctly we need to combine `__cplusplus` with `_MSVC_LANG` like we do in `compat/make_unique.h`
I noticed that there are several places where we use `__cplusplus` without `_MSVC_LANG`, so I will prepare another PR that fixes globally the detection of the language standard. However I think that the current PR is suitable for merging as it is now.

rbock (approved):

rbock:
Nice! I will wait for AppVeyor to complete and merge tomorrow.

rbock:
Thanks for the change! Merged.

PR #594, opened 2024-09-03
Fix C++ standard check on MSVC

MeanSquaredError:
If MSVC is run with the default options, the `__cplusplus` macro always has the value `199711L`, so sqlpp11's code that relies on that macro detects incorrectly the C++ standard version.

The proper way to detect the C++ standard version is first to check for the `_MSVC_LANG` macro and if it is present, use its value as the C++ standard version, otherwise use `__cplusplus`.
For details see [/Zc:__cplusplus](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170)

We already use this approach in `compat/make_unique.h`, by defining the custom macro `CXX_STD_VER`

This PR changes the code globally to use that custom macro `CXX_STD_VER`.
The PR was built and tested with
```
cmake --build build
cd build
ctest
```

### DIFF
--- a/include/sqlpp11/compat/cxx_std_ver.h
+++ b/include/sqlpp11/compat/cxx_std_ver.h
@@ -1,7 +1,0 @@
-#pragma once
-
-#ifdef _MSVC_LANG
-#define CXX_STD_VER _MSVC_LANG
-#else
-#define CXX_STD_VER __cplusplus
-#endif
--- a/include/sqlpp11/compat/cxx_std_ver.h
+++ b/include/sqlpp11/compat/cxx_std_ver.h
@@ -1,0 +1,7 @@
+#pragma once
+
+#ifdef _MSVC_LANG
+#define CXX_STD_VER _MSVC_LANG
+#else
+#define CXX_STD_VER __cplusplus
+#endif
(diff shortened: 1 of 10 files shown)


include/sqlpp11/compat/cxx_std_ver.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#ifdef _MSVC_LANG
+#define CXX_STD_VER _MSVC_LANG
```
CJCombrink, line 4: Is the define `CXX_STD_VER` specific to sqlpp11?
If yes I would rather advocate to rename it to be more specific to prevent any name conflicts with anyone else that might also have decided on this name.
Something like `SQLPP_CXX_STD_VER`?


_PS: I know this just moves the macro, but the question holds_

CJCombrink (approved):

include/sqlpp11/compat/cxx_std_ver.h
MeanSquaredError: Prefixing the macro makes sense. I renamed the macro: CXX_STD_VER -> SQLPP_CXX_STD

rbock:
Thank you both for the change!

PR #597, opened 2024-09-11
SQLite3: Enable execute() to run prepared_statements

margaretselzer:
Enable execute() to run prepared_statements for SQLite3.

Limit execute to run only non-select statements.
(Note: Does not apply to raw sql string select queries).

### DIFF
--- a/include/sqlpp11/sqlite3/connection.h
+++ b/include/sqlpp11/sqlite3/connection.h
@@ -374,12 +374,26 @@ namespace sqlpp
 
       template <
           typename Execute,
-          typename Enable = typename std::enable_if<not std::is_convertible<Execute, std::string>::value, void>::type>
+          typename std::enable_if<not std::is_convertible<Execute, std::string>::value 
+                                  and not sqlpp::is_prepared_statement_t<Execute>::value, int>::type = 0>
       size_t execute(const Execute& x)
       {
+        static_assert(not sqlpp::is_select_t<Execute>::value, "argument must not be a select statement - use operator() instead");
+
         _context_t context{*this};
         serialize(x, context);
         return execute(context.str());
+      }
+
+      template <
+          typename Execute,
+          typename std::enable_if<sqlpp::is_prepared_statement_t<Execute>::value, int>::type = 0>
+      size_t execute(const Execute& x)
+      {
+        static_assert(not sqlpp::is_select_t<Execute>::value, "argument must not be a select statement - use operator() instead");
+
+        operator()(x);
+        return static_cast<size_t>(sqlite3_changes(native_handle()));
       }
 
       template <typename Execute>
--- a/tests/sqlite3/usage/Execute.cpp
+++ b/tests/sqlite3/usage/Execute.cpp
@@ -23,6 +23,11 @@
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "Sample.h"
+#include <sqlpp11/alias_provider.h>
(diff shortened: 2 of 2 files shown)


margaretselzer:
If you think that limiting `execute()` to running only non-select queries is a good idea, then I have another suggestion as well. Instead of using `operator()()` to run select queries, there should be a `query()` function. 

Using `query()` then would signify that we expect returned (record) data while `execute()` would signify that we only want to run an operation where no records are returned. Similarly to `RetT function(...)` vs `void method(...)`.

IMHO using `operator()` obfuscates the intent - and the resulting code - and should be deprecated (if that was available in C++11 - but why do we still stick to C++11?).

WDYT?

include/sqlpp11/sqlite3/connection.h
```diff
@@ -372,14 +372,29 @@ namespace sqlpp
...
+        static_assert(not sqlpp::is_select_t<Execute>::value, "argument must not be a select statement - use operator() instead");
+
+        _context_t context{*this};
+        serialize(x, context);
+        return execute(context.str());
+      }
+
+      template <
+          typename Execute,
+          typename std::enable_if<not std::is_convertible<Execute, std::string>::value
```
rbock, line 390: Is this part of the condition necessary?

include/sqlpp11/sqlite3/connection.h
```diff
@@ -372,14 +372,29 @@ namespace sqlpp
...
-        _context_t context{*this};
-        serialize(x, context);
-        return execute(context.str());
+      template <
+          typename Execute,
+          typename std::enable_if<not std::is_convertible<Execute, std::string>::value 
+                                  and not sqlpp::is_prepared_statement_t<Execute>::value, int>::type = 0>
+      size_t execute(const Execute& x)
+      {
+        static_assert(not sqlpp::is_select_t<Execute>::value, "argument must not be a select statement - use operator() instead");
```
rbock, line 381: Wouldn't `SELECT FOR UPDATE` make sense to execute?

rbock (changes requested):
Looks pretty good to me already, see comments in the code.

include/sqlpp11/sqlite3/connection.h
margaretselzer: No, probably an overkill. Even if a prepared statement is convertible to string, we should prefer execution via this overload. I'll remove.

include/sqlpp11/sqlite3/connection.h
margaretselzer: I never used this so did not think of it. However, as it is a select that does return rows - even if as an extra function it also locks them - I do not think `execute()` should be used to running them. See my comment/question above on differentiation of `execute()` from `query()`.

Furthermore, AFAIK SQLite does not support this syntax. SQLite locks the whole database file for a write so it would not make much sense anyway.

rbock:
Thanks for the update. I forgot to ask for one more thing, sorry about the piece-meal review:

Can you add a test that uses the new form of `execute`?

rbock (approved):

rbock:
Approved, thanks for the change!

Thanks for your offer to continue on this path:

**MockDB**: Indeed, just add the new version of execute and add a test to make sure this compiles.

**MySQL and PostgreSQL**: If you can't test this yourself, then I will do it. No problem.

**query()**: I am fine with `query` being the primary function.

**Documentation**: Awesome question :-) I am very open to suggestions.

PR #598, opened 2024-09-11
Add argumentless overload for get_sql_name()

margaretselzer:
Convenience function that comes handy with template metaprogramming: no need to create an object of the type whose SQL name we desire to use.

Also works with non-default constructible objects.


rbock:
That looks reasonable to me, thanks!

PR #601, opened 2024-09-18
Enable execute() to run prepared_statements for all connectors

margaretselzer:
I am absolutely not sure about what applies to PostgreSQL and MySQL in terms of limitation to one query at a time or return value, but I hope that this is in any case a good baseline. 

I have taken the liberty to modify the return type of `execute()` for MySql from `void` to `size_t` so that it is in sync with other connectors.


rbock:
Thanks!

FYI: It might take a few days for me to take a closer look. I am a bit busy with something else right now :-)

Feel free to ping me if it feels like it takes too long.

rbock:
OK, this took much longer than it should have...

Thanks!

PR #602, opened 2024-09-23
Fix find_package_handle_standard_args

dg0yt:
Fixes first var being misunderstood as custom message.
Fixes actually checking the first var.

~~~
CMake Error at E:/vcpkg/downloads/tools/cmake-3.29.2-windows/cmake-3.29.2-windows-i386/share/cmake-3.29/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  MySQL_INCLUDE_DIR (missing: MySQL_LIBRARY)
~~~


rbock:
Looking at the documentation for `find_package_handle_standard_args` this makes sense to me.

Thanks for fixing!

PR #603, opened 2024-09-23
Update vcpkg section of README.md

dg0yt:
Don't advertise the connector ports. (Legacy.)
Indicate that a database feature must be selected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ git clone https://github.com/Microsoft/vcpkg.git
 cd vcpkg
 ./bootstrap-vcpkg.sh
 ./vcpkg integrate install
-vcpkg install sqlpp11[<database>]
+./vcpkg install 'sqlpp11[mysql]' # or other database feature
 ```
 
 The sqlpp11 port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -194,21 +194,16 @@ Some connectors can be installed with the formula. See `brew info marvin182/zapf
...
 You can download and install sqlpp11 using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
-   
+
 ```bash
 git clone https://github.com/Microsoft/vcpkg.git
 cd vcpkg
 ./bootstrap-vcpkg.sh
 ./vcpkg integrate install
-vcpkg install sqlpp11
+vcpkg install sqlpp11[<database>]
```
rbock, line 203: I have no clue about `vcpkg`, so apologies for the question: Is there maybe a `space` missing here?

i.e. `sqlpp11mysql` vs `sqlpp11 mysql`

dg0yt: No. The point is that users shall select an actual database "feature", e.g. `sqlpp11[mysql]`.

I understand that sqlpp11 is header-only, but from a package management perspective, the backends are transitive dependencies of sqlpp11.

Let's use an example instead, and resolve two additional quirks (path to vcpkg; zsh syntax):
```suggestion
./vcpkg install 'sqlpp11[mysql]' # or other database feature
```
Okay?

rbock:
Thanks for the explanation!

PR #608, opened 2024-12-06
Support binding results and parameters of type time_of_day with SQLite3

vaijns:
While testing with sqlite3 I got a compile-time error: "`class sqlpp::sqlite3::bind_result_t` has no member named `_bind_time_of_day_result`" and as I saw no reason to not have it, I've added it and it seems to work.

If there's anything else  to do, please let me know. C:

-Jonas

### DIFF
--- a/include/sqlpp11/sqlite3/bind_result.h
+++ b/include/sqlpp11/sqlite3/bind_result.h
@@ -204,6 +204,29 @@ namespace sqlpp
         }
       }
 
+      void _bind_time_of_day_result(size_t index, ::std::chrono::microseconds* value, bool* is_null)
+      {
+        if (_handle->debug)
+          std::cerr << "Sqlite3 debug: binding time_of_day result at index: " << index << std::endl;
+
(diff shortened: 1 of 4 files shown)


rbock:
Hi Jonas,

Thanks for adding this! Indeed, this was a pure oversight.

The change looks good to me.
Would you mind adding a test, too? It should be rather straight forward to add it to https://github.com/rbock/sqlpp11/blob/main/tests/sqlite3/usage/DateTime.cpp

Cheers,
Roland

vaijns:
Hi Roland,

Good thing you've mentioned adding tests... time_of_day parameters weren't supported either, which I should've checked before. 😅
I've added them to the other date and time tests (DateTime.cpp), hope that's okay and I also hope the test cases I've chosen make sense.

-Jonas

rbock:
Hi Jonas,

Great! Thanks for the update! This looks good to me.

Merging...

Cheers,
Roland

PR #609, opened 2024-12-06
Add SQL functions CURRENT_DATE, CURRENT_TIME and CURRENT_TIMESTAMP

vaijns:
I guess you can also just use std::chrono instead but with those functions (which are also standardized) it might feel more like SQL.

-Jonas

### DIFF
--- a/include/sqlpp11/current_date.h
+++ b/include/sqlpp11/current_date.h
@@ -1,0 +1,84 @@
+#pragma once
+
+/*
+ * Copyright (c) 2013-2015, Roland Bock
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification,
+ * are permitted provided that the following conditions are met:
+ *
+ *   Redistributions of source code must retain the above copyright notice, this
+ *   list of conditions and the following disclaimer.
+ *
+ *   Redistributions in binary form must reproduce the above copyright notice, this
+ *   list of conditions and the following disclaimer in the documentation and/or
+ *   other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR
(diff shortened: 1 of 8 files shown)


rbock:
Hi Jonas,

Thanks for adding these functions. Seems very reasonable to me.

Would be great if you could add serialization tests, e.g. similar to https://github.com/rbock/sqlpp11/blob/main/tests/core/serialize/Upper.cpp

Note: In the optional-no-dynamic branch (major overhaul), I decided to drop the `_auto_alias_t`. We might want to drop it for newly added functions before adding them. WDYT?

Cheers,
Roland

vaijns:
Hi Roland,

I've added some basic serialization tests, let me know if there's more to test but I guess the functions themselves are very basic as well, so I couldn't think of anything else.

I've mostly looked at the trim()-function, that's where the `_auto_alias_t` came from but I agree that it doesn't seem necessary.

-Jonas

On another note: I've thought about adding more date and time functions, namely `year_of`, `month_of`, ... (or maybe `extract` instead of those), `datediff`, `date_add`, `date_sub`. Is that something you'd be willing to merge or do you think this should rather be done in application code? One problem with this is not being standardized functions of course.
If you're interested I'd create an issue for it to discuss design details.

include/sqlpp11/current_date.h
```diff
@@ -0,0 +1,79 @@
...
+          return current_date;
+        }
+        const T& operator()() const
+        {
+          return current_date;
+        }
+      };
+    };
+
+    constexpr current_date_t()
```
rbock, line 60: Can you make this `default`?

include/sqlpp11/current_date.h
```diff
@@ -0,0 +1,79 @@
...
+namespace sqlpp
+{
+  struct current_date_t : public expression_operators<current_date_t, day_point>, public alias_operators<current_date_t>
+  {
+    using _traits = make_traits<day_point, tag::is_expression, tag::is_selectable>;
+
+    using _nodes = detail::type_vector<>;
+    using _can_be_null = std::false_type;
+
+    struct _alias_t
```
rbock, line 41: Do you actually need this? I would assume that you only need this for `select(sqlpp::current_date)`?

In the next generation of the library, I would (most likely) say this has to `select(sqlpp::current_date.as(some_alias))`.

One of the reasons for this is that I am generally removing some implicit logic (including implicit names). Another reason is code complexity. Automatic names make things harder...

rbock (commented):
Thanks for the update including the tests! Added two more comments, that apply to all functions.

rbock:
Regarding the proposed additional functions: Please open a separate issue.

include/sqlpp11/current_date.h
vaijns: Requiring explicit aliases seems reasonable. I didn't really know what/where `_nodes` and `_can_be_null` are used for and didn't know they were optional. Removed them now.

rbock:
Hi Jonas,

Thanks for the cleanup. This looks good to me.

Merging...

Cheers,
Roland

Issue #621, opened 2025-05-20
GCC compilation fails when using -Werror=undef


wgraef:
When using compilation option `-Werror-undef` with GCC, compilation fails due to unknown `_MSC_FULL_VER` in `include/sqlpp11/chrono.h` and `tests/core/usage/DateTime.cpp`.

First check whether the variable exists, before evaluating its value.

rbock:
This should be fixed with [5e930f3](https://github.com/rbock/sqlpp11/commit/5e930f3901c68c1900873a42c972deed322fd08c)

PR #622, opened 2025-05-20
Allow undefined _MSC_FULL_VER

wgraef:
See #621

### DIFF
--- a/include/sqlpp11/chrono.h
+++ b/include/sqlpp11/chrono.h
@@ -37,7 +37,7 @@ namespace sqlpp
     using day_point = std::chrono::time_point<std::chrono::system_clock, days>;
     using microsecond_point = std::chrono::time_point<std::chrono::system_clock, std::chrono::microseconds>;
 
-#if _MSC_FULL_VER >= 190023918
+#if (!defined _MSC_FULL_VER) || _MSC_FULL_VER >= 190023918
     // MSVC Update 2 provides floor, ceil, round, abs in chrono (which is C++17 only...)
     using ::std::chrono::floor;
 #else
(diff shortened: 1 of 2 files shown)


include/sqlpp11/chrono.h
```diff
@@ -37,7 +37,7 @@ namespace sqlpp
     using day_point = std::chrono::time_point<std::chrono::system_clock, days>;
     using microsecond_point = std::chrono::time_point<std::chrono::system_clock, std::chrono::microseconds>;
 
-#if _MSC_FULL_VER >= 190023918
+#if (!defined _MSC_FULL_VER) || _MSC_FULL_VER >= 190023918
```
rbock, line 40: This would break for C++14 and below since they don't have std::chrono.

So it should probably be this?

```c++
#if (defined _MSC_FULL_VER) and _MSC_FULL_VER >= 190023918
```

wgraef: Good point, but then std::chrono would never be used by GCC, so how about
```
#if (defined _MSC_FULL_VER && _MSC_FULL_VER >= 190023918) || __cplusplus >= 201703L
```

rbock: That would work, happy to accept.

But it also seems like a micro optimization given that `date::` is used in quite a few places in the library. If using `std::chrono` instead of `data` is what you are strive for, then a few more changes would be required.

Alternatively, [sqlpp23](https://github.com/rbock/sqlpp23) might be an option?

rbock:
@wgraef Just to make sure we're not waiting on each other:

I am fine with your proposed solution

```c++
#if (defined _MSC_FULL_VER && _MSC_FULL_VER >= 190023918) || __cplusplus >= 201703L
```

wgraef:
Thanks, I will also look into sqlpp23.

rbock:
Done in [5e930f3](https://github.com/rbock/sqlpp11/commit/5e930f3901c68c1900873a42c972deed322fd08c)

PR #630, opened 2025-12-12
Fix compile warning (shadowed variable)

lbatuska:
Fixed a variable shadowing issue causing a warning with `-Wshadow`. Renamed the password constructor parameter to avoid conflicting with the class member.


rbock:
Thanks!

